PR #63, opened 2020-01-06
Decrease bundle size by excluding BIP39 non-english json files

pradel:


pradel:
@Sirlupinwatson1 thanks for the feedback!

I tried and didn't notice any error, but I will try to test more deeply to see if I can see any edge case :)

Sirlupinwatson1:
> @Sirlupinwatson1 thanks for the feedback!
> 
> I tried and didn't notice any error, but I will try to test more deeply to see if I can see any edge case :)

Great job. 

I did review it and only found warnings, no error related.

pradel:
Closing this one as the SDK changed to stacks.js I will do the same in a separate pr

PR #149, opened 2020-11-22
feat: new editor for markdown

pradel:


pradel:
Decided to not go with makdown approach

PR #229, opened 2021-09-18
feat: setup new design system and redesign login page

pradel:


sigle/.babelrc
```diff
@@ -1,4 +1,4 @@
 {
   "presets": ["next/babel"],
-  "plugins": ["babel-plugin-macros", ["styled-components", { "ssr": true }]]
+  "plugins": ["babel-plugin-macros"]
```
pradel, line 3: this will likely create some issues but using radix and this at the same time is not working 😀

pradel: YOLO

sigle/src/pages/login.tsx
```diff
@@ -98,44 +63,59 @@ const Login = () => {
...
+                width={100}
+                height={44}
+              />
+            </a>
+            <Heading as="h1" size="2xl" css={{ mt: '$15' }}>
+              Welcome!
+            </Heading>
+            <Text css={{ mt: '$7' }}>
+              Sigle is a web 3.0 open source blogging platform focused on{' '}
+              <strong>protecting your privacy</strong> built on top of Stacks.
```
Akirtovskis, line 83: @pradel *and* build on top of Stacks

Akirtovskis (approved):

PR #283, opened 2022-01-17
feat: smart error page

gregogun:
### Description

This PR does the following: 

- Conditionally renders error page content based on the type of error and whether user is logged in or not
- Replaces tailwindcss and styled components with design system 
- Responsive across varying device sizes
- Adds xs breakpoint for better management of illustrations on small screen sizes
- Adds new header component that displays content based on whether a user is logged in or not
- (500 page) Refresh button reloads page

### TODO
- Add test to assert that an Sentry ID renders when there is a server error

### Current Issues
- I'm currently struggling to obtain the Sentry ID from the GetInitialProps (tested it by storing `Sentry.captureException(err)` in a variable and passing it to error page via props, using useEffect to print value and it always returns undefined) 

Any feedback or help with the issues will be appreciated!


sigle/src/pages-lib/api/feed/[username].test.ts
```diff
@@ -39,7 +39,7 @@ describe('test feed api', () => {
     expect(res.end).toBeCalled();
     const jsonObj = parser.parse(res.end.mock.calls[0][0]);
     expect(jsonObj.rss.channel).toEqual({
-      copyright: 'All rights reserved 2021, sigleapp.id.blockstack',
+      copyright: 'All rights reserved 2022, sigleapp.id.blockstack',
```
Comment (pradel) [inline sigle/src/pages-lib/api/feed/[username].test.ts:42]: Can you merge the latest `main` branch in your pr? This should resolve the issues.

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,38 +1,136 @@
...
-import { sigleConfig } from '../config';
+import { Button, Text } from '../ui';
+import { Container } from '../ui';
+import Image from 'next/image';
+import { useAuth } from '../modules/auth/AuthContext';
 
-const NotFoundContainer = styled(Container)`
-  ${tw`mt-8 mb-8 flex flex-col items-center text-center md:flex-row md:text-left`};
-`;
+const NotFoundContainer = styled(Container, {
```
pradel, line 13: Right now the content is just below the header, what do you think about having it centered in the bottom block? Similar to how the Login page is behaving (without the header for the login but same idea). Should be possible to do it easily using flex.

Capture d’écran 2022-01-18 à 13 12 42

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Button as ButtonBase, Container, Flex, Text } from '../../../ui';
+import { useAuth } from '../../auth/AuthContext';
+import { Logo } from './Logo';
+
+const Header = styled('header', Container, {
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  mt: '$10',
+  mb: '$20',
```
pradel, line 14: Regarding the header, do we have the same bottom spacing in all the pages? If not might be better to have the same mt and mb value here and adjust individually on the different pages.

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+});
+
+const StatusDot = styled('div', {
+  backgroundColor: '#37C391',
+  width: '$2',
+  height: '$2',
+  borderRadius: '$round',
+});
+
+const Button = styled('a', ButtonBase);
```
pradel, line 32: When we don't add any style, I tend to prefer using the `as` property to change the element easily, also make it more clear that the button is a link by just reading the JSX part of the file.

```
<Button as={'a'}>
 ....
</Button>
```

In the same idea, line 17 for the nav could be simplified to:

```
<Flex as="nav" gap="10">
 ....
</Flex>
```

Wdyt of this approach?

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+  return (
+    <Header>
+      <Nav align="center">
+        <Link href="/" passHref>
+          <a>
+            <Logo />
+          </a>
+        </Link>
+        {user && (
+          <Link href="/" passHref>
```
pradel, line 51: Clicking on the link should open the blog in a new tab :)

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+          </Link>
+        )}
+      </Nav>
+      {user && (
+        <Flex align="center" gap="10">
+          <Flex gap="1" align="center">
+            <StatusDot />
+            <Text>{user.username}</Text>
+          </Flex>
+          <Link href="/" passHref>
```
pradel, line 67: Should be linked to the current logic of creating a new story :)

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,74 @@
...
+  if (router.pathname === '/login') {
+    return null;
+  }
+
+  return (
+    <Header>
+      <Nav align="center">
+        <Link href="/" passHref>
+          <a>
+            <Logo />
```
pradel, line 47: Is there a reason for the logo to be an SVG and not use the next `Image` component?

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,38 +1,136 @@
...
+    alignItems: 'start',
+  },
+
+  px: 0,
+  pr: '$5',
+  display: 'flex',
+  flexDirection: 'column',
+});
+
+const NotFoundTitle = styled('h1', {
```
pradel, line 52: The titles are not following the design system? Is there a way to reuse the Heading components we have?

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,38 +1,136 @@
...
+
+const NotFoundIllu = ({
+  statusCode,
+  alt,
+}: {
+  statusCode: number;
+  alt: string;
+}) => (
+  <Image
+    src={`/static/img/${statusCode === 404 ? '404' : '5XX'}.png`}
```
pradel, line 124: Are dynamic src values on the Image component allowed in the Next.js? I thought this was forbidden so the image paths can be extracted at build time. Not 100% sure about this tbh and I am not able to find a clear answer in the documentation.

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,38 +1,136 @@
...
+  alignItems: 'center',
+  mx: 'auto',
+  mt: '$8',
+  mb: '$3',
+});
 
-const NotFoundSubTitle = styled.div`
-  ${tw`mb-8 text-3xl`};
-`;
+const NotFoundTextContainer = styled(Container, {
```
pradel, line 33: Is there a reason this one is extending the `Container` instead of just being a `div`? I know it was like this before but probably just my mistake at this time :p

pradel (commented):
A couple of comments regarding the way we organise things in the code (so we are on the same page for changes in the future), otherwise the page looks great, good job 🚀  Let's have a discussion regarding how to handle Sentry in a call.

sigle/src/pages-lib/api/feed/[username].test.ts
Comment (gregogun) [reply sigle/src/pages-lib/api/feed/[username].test.ts:42]: Sure. Will I need to revert the commit/undo the change or will it be fine to merge main into the code as it is?

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: Yes after your explanation and seeing the examples, it definitely makes more sense to use this approach!

sigle/src/pages-lib/api/feed/[username].test.ts
Comment (pradel) [reply sigle/src/pages-lib/api/feed/[username].test.ts:42]: You can merge the code as it is, and resolve the conflicts is there are any :)

gregogun:
@QuentinSaubadu 

Here is the link to the live preview: https://sigle-7h7akhy33-leopradel.vercel.app/404

QuentinSaubadu:
> @QuentinSaubadu
> 
> Here is the link to the live preview: https://sigle-7h7akhy33-leopradel.vercel.app/404

- Title Line height (font-size-9) : 72px

-   H3 is 24px (instead of 21) 
-   H3 Line Height : 30px (instead of default)  
  
-   "Action" font size (all header fonts) should be 14px (like you have on the CTA) instead of 16, but I’ll move it to 15px. 14 is finally too small.
  
      - Logo is not aligned vertically   
Capture d’écran 2022-01-21 à 08 57 02

    
    - The CTA  is not visible on mobile (you need to scroll to view it) is a bit annoying me.. It is my bad 100% I didn’t do the mobile version of it, I’m sorry for that.  Can you reduce the margin between header and image, reduce the image just a bit also so we can see the CTA?  
    
-Add 40px margin under CTA on mobile version

Thank you Greg

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -74,28 +80,47 @@ export const AppHeader = () => {
 
   return (
     <Header>
-      <Flex gap="10" as="nav" align="center">
+      <Flex justify="center" gap="10" as="nav" align="center">
         <Link href="/" passHref>
-          <a>
+          <a style={{ height: '29px' }}>
```
gregogun, line 85: After inspecting via dev tools, the issue with the alignment between the logo and "visit blog" button seemed to be caused by some default height on the `a` tag. Forcing a height on the `a` tag seems to be the only way to currently achieve alignment.

Let me know if you have any thoughts on how to better solve this problem!

sigle/src/styles/index.css
```diff
@@ -1 +1,8 @@
 @tailwind base;
+
+html,
+body,
+#root,
+#__next {
+  height: 100%;
```
gregogun, line 7: Context: after trying other solutions to implement centering the content below the header, this one allowed me to do so. 

Not sure if there is a desired place for global styles like this, but I'm aware of the`globalCss` function from stitches, so happy to put them there if preferred.

Image from suggested change:
Capture d’écran 2022-01-18 à 13 12 42

sigle/src/modules/layout/components/AppHeader.tsx
pradel: Just adding diplay flex to the `a` tag is working )

sigle/src/stitches.config.ts
```diff
@@ -94,6 +94,7 @@ export const { styled, css, globalCss, keyframes, getCssText, createTheme } =
       },
     },
     media: {
+      xs: '(min-width: 280px)',
```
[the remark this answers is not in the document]
gregogun: I've just tried with the recommended approach, and in doing so I've remembered the context of why I added the xs:

It seems that when adding default values and media queries, the default values seem to take precedence. Here's what is happening when I remove xs and add the width and height as base styles: 
Screenshot 2022-01-24 at 12 47 57




I remember having issues with this in other projects, but perhaps I may be implementing this incorrectly. Have you experienced this at all when working with media queries in stitches? 

The code: 
Screenshot 2022-01-24 at 12 56 31

sigle/src/pages/_app.tsx
```diff
@@ -61,6 +61,17 @@ Router.events.on('routeChangeComplete', () => {
...
 
 const GlobalStyle = createGlobalStyle`
+
+  html, body, #root, #__next{
+    height: 100%;
+  }
+
+  #__next {
+    width: 100%;
+    position: absolute;
```
pradel, line 71: can we remove this part? doesn't seem like a really good pattern to do this on the main component.

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,218 @@
...
+    );
   }
-`;
+
+  return (
+    <Image
+      src="/static/img/5XX.png"
+      alt={alt}
+      priority
+      layout="fill"
```
pradel, line 123: Looks like using "fill" in the next image component is creating some issues with the responsiveness on some screens. Using either the default layout or the responsive one should fix this issue :)

Screenshot 2022-01-25 at 14 16 42

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,135 @@
...
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  mt: '$4',
+
+  '@md': {
+    mt: '$10',
+  },
+});
+
+const Logo = () => (
```
pradel, line 31: as the component is really simple, it might be better to just inline it in the code, wdyt?

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -0,0 +1,135 @@
...
+            </Text>
+            <EyeOpenIcon />
+          </Button>
+        )}
+      </Flex>
+      {user && (
+        <Flex
+          css={{
+            display: 'none',
+            '@xl': {
```
pradel, line 110: question for you and Quentin but would the `md` breakpoint still work at this step?

sigle/src/pages-lib/_error.test.tsx
```diff
@@ -94,3 +94,18 @@ describe('getInitialProps', () => {
     expect(res).toEqual({ hasGetInitialPropsRun: true, statusCode: 404 });
   });
 });
+
+describe('Sentry ID', () => {
+  it('should display server error Sentry ID', async () => {
```
pradel, line 99: 👌

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,218 @@
...
-  ${tw`mt-8 mb-8 flex flex-col items-center text-center md:flex-row md:text-left`};
-`;
+const NotFoundContainer = styled('div', {
+  boxSizing: 'border-box',
+  display: 'flex',
+  flexDirection: 'column',
+  justifyContent: 'center',
+  alignItems: 'center',
+  textAlign: 'center',
+  height: '90%',
```
pradel, line 19: using an height at 90% here sounds like something that will not adapt with some screen size, if the height is not big enough I can still see a small scroll in the page

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,218 @@
...
+  my: 'auto',
+  mx: 'auto',
 
-const NotFoundTextContainer = styled(Container)`
-  ${tw`flex flex-col items-center md:items-start md:pl-8`};
-`;
+  '@lg': {
+    width: '90%',
```
pradel, line 24: What is the reason to not have a full width around this breakpoint?

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,218 @@
...
+            <NotFoundIllu statusCode={404} alt="A lost traveller" />
+          </NotFoundWrapper>
+        ) : (
+          <ErrorWrapper>
+            <NotFoundIllu statusCode={500} alt="Woodpecker and broken pencil" />
+          </ErrorWrapper>
+        )}
+        <NotFoundTextContainer>
+          <Heading
+            css={{ mb: '$3', lineHeight: 1 }}
```
pradel, line 170: why is the line-height different, is it an issue with the design of the page or the values of the current design system?

pradel (commented):
The new logo needs to be deleted

sigle/src/pages-lib/_error.tsx
gregogun: I'm currently getting the following error when trying to add any other layout than "fill":

Screenshot 2022-01-25 at 15 24 42
 
Not sure how to work around  as we can't add a responsive width/height directly on Image component - alternatively, I could use the img component but will of course lose benefits of the Image component. What are your thoughts?

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,214 @@
...
+import Image from 'next/image';
+import { useAuth } from '../modules/auth/AuthContext';
+import { AppHeader } from '../modules/layout/components/AppHeader';
 
-const NotFoundContainer = styled(Container)`
-  ${tw`mt-8 mb-8 flex flex-col items-center text-center md:flex-row md:text-left`};
-`;
+const NotFoundContainer = styled(Container, {
+  position: 'relative',
+  boxSizing: 'border-box',
```
pradel, line 14: These 2 props can be removed, they are not changing anything :)

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,214 @@
...
+    height: '403px',
+  },
+
+  '@2xl': {
+    width: '765px',
+    height: '593px',
+  },
+});
+
+const NotFoundIllu = ({
```
pradel, line 91: Can you inline this component please?

sigle/src/pages-lib/_error.tsx
```diff
@@ -1,77 +1,214 @@
...
+const NotFoundContainer = styled(Container, {
+  position: 'relative',
+  boxSizing: 'border-box',
+  display: 'flex',
+  flexDirection: 'column',
+  justifyContent: 'center',
+  alignItems: 'center',
+  textAlign: 'center',
+  my: 'auto',
+  mx: 'auto',
```
pradel, line 21:
```
 my: 'auto',
  mx: 'auto',
```
these props are already injected by `Container`, it's safe to remove them

pradel (commented):
Some last nitpick and good to go 🚀

pradel (approved):

PR #301, opened 2022-01-27
fix: header issue

gregogun:


sigle/src/modules/editor/components/SlateEditor.tsx
```diff
@@ -619,6 +623,18 @@ export const SlateEditor = ({
             </Box>
             <span>{story.type === 'public' ? ' | Published' : ' | Draft'}</span>
           </Text>
+          {story.type === 'public' && (
+            <Button
+              css={{ display: 'flex', alignItems: 'center', gap: '$2' }}
```
gregogun, line 628: although I'm centering the items here, it seems that the eye icon visually is not properly centered. Are you seeing the same thing?

PR #307, opened 2022-02-05
feat: analytics API

pradel:
For now just playing around to see what's possible to do with the Fathom API


gregogun (approved):
awesome! response format looks good to me 👍

PR #317, opened 2022-02-18
feat: link new editor to edit page under a flag

pradel:
Steps to use the new Editor:
- Create a new draft
- Add `?experimentalEditor=true` at the end of the URL to show the new editor


sigle/src/modules/editor/EditorHeader.tsx
```diff
@@ -0,0 +1,95 @@
...
+interface EditorHeaderProps {
+  story: Story;
+  loadingSave: boolean;
+  onPublish: () => void;
+  onUnpublish: () => void;
+  onOpenSettings: () => void;
+  onSave: () => void;
+}
+
+export const EditorHeader = ({
```
pradel, line 21: moved this to a new component so both old and new editor have access to it

Akirtovskis (approved):

Issue #270, opened 2022-01-03
Upgrade tailwindcss typography package to v5


pradel:
## Description

Upgrade the [tailwindcss typography](https://github.com/tailwindlabs/tailwindcss-typography/releases/tag/v0.5.0) plugin to the latest version.

We need to investigate if an upgrade to tailwind 3 is required first.

PR #319, opened 2022-02-19
feat: create separate tailwind package

pradel:
The goal of this pr is to create a separate package that can be used by all clients rendering Sigle content. This also mean we can remove any tailwind dependency on the frontend next.js app when the redesign is done.
We use tailwind CSS customised to create a CSS file used to style the content.
It contains:
- reset CSS 
- typography plugin rules

We changed all the font-size, letter-spacing, line-height, padding of all the elements. Images are also centered by default now. The new reading experience is much better :D

Note: the old editor won't be affected by this change for now, you will see the changes only on the new editor.

Fixes: https://github.com/sigle/sigle/issues/270


sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -61,10 +61,6 @@ const StyledEditorContent = styled(EditorContent, {
   '& img.ProseMirror-selectednode': {
     outline: '1px solid $orange11',
   },
-  // Code block style
-  '& pre': {
```
pradel, line 65: migrated this to the tailwind package

Akirtovskis (approved):

gregogun (approved):

PR #321, opened 2022-02-21
feat: redesign settings modal

gregogun:


sigle/src/ui/Form.ts
```diff
@@ -0,0 +1,89 @@
...
+      boxShadow: '0 0 0 2px $colors$gray8',
+    },
+
+    '&::placeholder': {
+      color: '$gray9',
+    },
+  },
+
+  '&[type="date"]::-webkit-calendar-picker-indicator': {
+    background: 'url(/static/img/Calendar.svg) no-repeat',
```
pradel, line 44: You used an SVG as it was not possible to import a radix icon from the npm package, correct?

sigle/src/modules/editor/components/StorySettings.tsx
```diff
@@ -57,10 +50,21 @@ interface Props {
 export const StorySettings = ({ open, onClose, story, onSave }: Props) => {
   return (
     <Dialog open={open} onOpenChange={onClose}>
-      <StyledDialogContent aria-label="Story settings">
+      <StyledDialogContent
+        aria-label="Story settings"
+        css={{
```
pradel, line 55: As this is already a styled content you can move the `css` prop values to `styled(DialogPrimitive.Content, {`

sigle/src/ui/IconButton.ts
```diff
@@ -6,8 +6,16 @@ export const IconButton = styled('button', {
   alignItems: 'center',
   fontSize: '$2',
 
+  '& svg': {
+    color: '$gray11',
+  },
+
   variants: {
     size: {
+      sm: {
```
pradel, line 15: sm seems to have the same values as md here 🤔

sigle/src/ui/IconButton.ts
```diff
@@ -6,8 +6,16 @@ export const IconButton = styled('button', {
   alignItems: 'center',
   fontSize: '$2',
 
+  '& svg': {
```
pradel, line 9: Can you please explain what was not working before?

sigle/src/ui/Form.ts
```diff
@@ -0,0 +1,89 @@
+import { styled } from '../stitches.config';
+
+export const FormRow = styled('div', {
```
pradel, line 3: as this is not yet part of the design system I am wondering if there should be in the UI folder or not yet as we don't know the final version.
Wdyt?

pradel (changes requested):
Nice one, found the following issues so far while testing:
1. overlay does not have anymore an animation, was better before I think
2. The Save button is not shown, I have to hover it to see it
Screenshot 2022-02-23 at 08 24 36
 3. Meta description width is larger than a normal input

sigle/src/ui/Form.ts
gregogun: Yes, that's correct!

pradel:
Just wanted to add that for comment 2 "The Save button is not shown, I have to hover it to see it" it was a general issue which was solved yesterday, so merging main in your branch should solve the issue

sigle/src/ui/IconButton.ts
gregogun: Oh I think the value for sm was actually meant to be `$1` for padding, but after looking at the design system I can see there is only one size for the Icon Button with padding of `$1` so will remove the sm and add that change.

sigle/src/ui/IconButton.ts
gregogun: I think here I was testing to confirm the color would change, as it seemed not to work with `css` prop, but I think this was actually because the page hadn't refreshed - will remove this!

sigle/src/ui/Form.ts
gregogun: yes that's a good point, I'll move it to `StorySettingsForm.tsx` ?

pradel: Sounds good 👍

pradel: Can you also create an issue so we don't forget to create the components later please?

gregogun: Sure, will do that 👍

PR #326, opened 2022-02-21
feat: new editor publish story

pradel:
Allow stories to be published / unpublished in the new editor

Test cases:
1. on the current editor try to publish then unpublish to verify everything is still working correctly with the refactor
2. on the new editor try to publish then unpublish


Akirtovskis:
@pradel 
- old editor everything worked as expected.
- new editor publish and unpublish didn't seem to work on the UI side ( checked network tab and there seemed to be no problems )

Akirtovskis (approved):

PR #327, opened 2022-02-22
feat: new editor public view

pradel:
Use TipTap to render the public view of the articles with the correct style.


sigle/src/modules/publicStory/components/PublicStory.tsx
```diff
@@ -173,6 +174,10 @@ const CustomStyle = createGlobalStyle<{ siteColor?: string }>`
     `}
 `;
 
+const DynamicTipTapEditor = dynamic<{ story: Story; editable: boolean }>(() =>
```
pradel, line 177: we dynamically load the editor so it won't be loaded on pages using the old slate editor content, that way we don't ship a big bundle size on all pages

Akirtovskis (approved):

PR #355, opened 2022-02-28
feat: redesign article page

### LINKED ISSUE #281: Redesign public Story page
## Description

Screenshot 2022-02-28 at 09 09 25

gregogun:
This pr adds a redesign for the public story page, as well as updating the header for logged out users.

Resolves #281


sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -74,10 +81,36 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
...
+                }}
+              >
+                {format(story.createdAt, 'MMM dd')}
+                <span>•</span>
+                <span>{storyReadingTime?.text}</span>
+              </Text>
+            </Box>
+          </a>
+        </Link>
+        <Heading size="3xl">{story.title}</Heading>
```
pradel, line 112: the current H1 should already have the proper styling, we can leave it that way :) (style is coming from tailwind)

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -100,33 +122,50 @@ export const AppHeader = () => {
...
-            <Button onClick={handleCreateNewPrivateStory} size="lg">
-              Write a story
-            </Button>
-          )}
-          {loadingCreate && (
-            <Button disabled size="lg">
-              Creating new story...
+        ) : (
+          <Flex gap="6">
+            <a href={sigleConfig.twitterUrl} target="_blank" rel="noreferrer">
```
pradel, line 142: should we use IconButton here?

sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -74,10 +81,36 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
...
+            p: 0,
+          },
         }}
       >
-        <h1 className="sigle-title">{story.title}</h1>
-        {story.coverImage && (
+        <Link href="/[username]" as={`/${username}`} passHref>
+          <a>
+            <Text size="action">{siteName}</Text>
+            <Box css={{ mt: '$1', mb: '$7' }}>
```
pradel, line 96: is the box necessary around the Text? probably the margin can be added to the text to avoid using box :)

sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -74,10 +81,36 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
           "& :where(a strong):not(:where([class~='not-prose'] *))": {
             color: safeSiteColor,
           },
+
+          px: '$4',
+
+          '& p': {
```
pradel, line 87: why was this added?

sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -74,10 +81,36 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
           "& :where(a strong):not(:where([class~='not-prose'] *))": {
             color: safeSiteColor,
           },
+
+          px: '$4',
```
pradel, line 85: PublicStoryContainer is already defining padding line 18, should this be moved there?

sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -43,6 +46,8 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
   const seoDescription = story.metaDescription;
   const seoImage = story.metaImage || story.coverImage;
 
+  const showCoverImage = story.coverImage && !story.hideCoverImage;
```
pradel, line 49: we are deprecating the usage of hideCoverImage, can you please revert to how it was before?

sigle/src/modules/publicStory/PublicStory.tsx
pradel: if it's to remove the padding around the `p` elements you can use the `className="not-prose"` which will remove all the styles

gregogun: ohh I see, will do that instead - thanks!

sigle/src/icons/Discord.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
```
pradel, line 1: @gregogun just saw `@radix-ui/react-icons` published a new release, can you update the package and see if it's working to import it from there?

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -42,8 +48,11 @@ const StatusDot = styled('div', {
 
 export const AppHeader = () => {
   const { user } = useAuth();
-  const router = useRouter();
+  const router = useRouter() || { query: { username: '' } };
```
pradel, line 51: I think here it would be better to create a global mock like this https://github.com/vercel/next.js/issues/7479#issuecomment-533657701.

That way all the future tests would benefit from the fix too. Also the code itself should not contain special code related to the tests :)

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -68,8 +77,20 @@ export const AppHeader = () => {
   return (
     <Header>
       <Flex justify="center" gap="10" as="nav" align="center">
+        <Link href="/[username]" as={`/${username}`} passHref>
```
pradel, line 80: there is no pointer on hover right now, can please add it?

sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -90,6 +115,7 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
               css={{
                 marginLeft: '-$20',
                 marginRight: '-$20',
+                width: '100%',
```
pradel, line 118: what was the reason to add this one? Adding it make the negative margin not work properly

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: Ah I see! That makes sense. Where would be the best place to add the global mock?

pradel: You can do like this https://jestjs.io/docs/manual-mocks#mocking-node-modules

sigle/src/modules/publicStory/PublicStory.tsx
gregogun: yes so after speaking with Quentin regarding the design, he wanted the max width of the images to be the size of the content width, so that's why the full width property is there. I can check back with him regarding this? :)

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: Great, thanks :)

PR #357, opened 2022-03-01
fix: fix editor title disappearing

pradel:


sigle/src/modules/editor/CoverImage.tsx
```diff
@@ -29,28 +29,31 @@ export const CoverImage = ({ story, setStoryFile }: CoverImageProps) => {
...
+        contentType: blob.type,
+      });
 
-    setLoadingSave(false);
-    setStoryFile({ ...story, coverImage: coverImageUrl });
-  }, []);
+      setLoadingSave(false);
+      setStoryFile({ ...story, coverImage: coverImageUrl });
+    },
+    [story]
```
pradel, line 55: only line that really changed is this one, the story reference was not updated

gregogun (approved):

PR #370, opened 2022-03-07
feat: editor kbd shortcuts

### LINKED ISSUE #276: ⚡ Feature | Editor keyboard shortcuts
## Description

On the bottom right corner of the editor page the user can click on the keyboard and a list of available shortcuts should be visible.

Capture d’écran 2022-01-17 à 09 55 29

> @pradel to create the list of available shortcut for the new Editor

The current editor shortcuts are listed here https://docs.sigle.io/editing/keyboard-shortcuts

gregogun:
This pr adds the dialog for the editor shortcuts and hints. 

Resources: 
- https://dev.to/jasonwebb/how-to-build-a-more-accessible-carousel-or-slider-35lp

Resolves #276


sigle/src/ui/Dialog.tsx
```diff
@@ -63,11 +63,13 @@ const StyledCloseButton = styled(DialogPrimitive.Close, {
   right: '$6',
 });
 
-type DialogContentProps = React.ComponentProps<typeof DialogPrimitive.Content>;
+type DialogContentProps = React.ComponentProps<
+  typeof DialogPrimitive.Content
+> & { css?: CSS };
```
gregogun, line 68: allow shortcuts/hints instance of the dialog to control its width

sigle/src/stitches.config.ts
```diff
@@ -61,132 +61,141 @@ const orangeDark = {
...
+});
 
 export const darkTheme = createTheme('dark-theme', {
   colors: {
     ...grayDark,
     ...orangeDark,
   },
 });
+
+export type CSS = Stitches.CSS<typeof config>;
```
gregogun, line 202: get typings for css prop

sigle/src/modules/editor/ShortcutsDialog.tsx
```diff
@@ -96,7 +94,7 @@ export const ShortcutsDialog = () => {
             <TabsTrigger value="hints">Hints</TabsTrigger>
           </TabsList>
 
-          <TabsContent css={{ height: 650 }} value="shortcuts">
+          <TabsContent css={{ height: 600 }} value="shortcuts">
```
gregogun, line 97: fixed height here is to stop the height adjusting for each shortcuts tab creating undesirable UX - not sure if this is the best solution but works for now

sigle/src/modules/editor/ShortcutsDialog.tsx
```diff
@@ -0,0 +1,448 @@
...
+                </TabsTrigger>
+                <TabsTrigger value="p-formatting">
+                  Paragraph Formatting
+                </TabsTrigger>
+                <TabsTrigger value="selection">Text Selection</TabsTrigger>
+              </TabsList>
+              <TabsContent value="essentials">
+                <Table>
+                  <Tr>
+                    <Th scope="row">Copy</Th>
```
pradel, line 112: I feel like all this part could be simplified a lot. By doing a javascript object, adding new items later will be easier too :)

smth like :
```tsx
const array = [{
  action: 'Copy',
  command: '...'
}, // ....rest
];

array.map(() => (
<Tr>
<Th scope="row">{array.action}</Th>
// ..... rest
))
```

sigle/src/modules/editor/HintsCarousel.tsx
```diff
@@ -0,0 +1,151 @@
...
+  width: 600,
+  height: 350,
+  position: 'relative',
+  mx: 'auto',
+  boxShadow: '0 0 0 1px $colors$gray7',
+  borderRadius: '$1',
+  overflow: 'hidden',
+});
+
+export const HintsCarousel = () => {
```
pradel, line 23: I was wondering, can we display the content only of the current selected tab? That way the GIF will be loaded only when you display the hint, also it means it starts only when you navigate to it and does not continue to play in the background. Wdyt?

pradel (commented):
Good job on this one, some minor comments to improve the code otherwise good to go!

Can you just move the files to a subfolder? eg: `modules/editor/EditorShortcuts`

sigle/src/modules/editor/ShortcutsDialog.tsx
gregogun: Ahh yes of course, this makes much more sense 🤦‍♂️  Thanks for the suggestion! 

Do you think it would best to store the array in its own config file?

pradel: As it's the only place where we use it it's fine to have it in the same file

sigle/src/modules/editor/HintsCarousel.tsx
gregogun: I just checked in dev tools when the images are loading, and wondering if this is what we want to happen? (GIF loads only when we click on Tab) or does it continue playing after we change the Tab and we need to prevent that?

https://user-images.githubusercontent.com/65421744/157079873-22fcdf5b-fc09-4a72-969b-81c8ddd5e9ba.mov

pradel: Okay indeed I was confused as it continued playing in the background, if it starts when you load the tab then all good 👍

sigle/src/modules/editor/EditorShortcuts/ShortcutsDialog.tsx
```diff
@@ -0,0 +1,340 @@
...
+                </TabsTrigger>
+                <TabsTrigger value="p-formatting">
+                  Paragraph Formatting
+                </TabsTrigger>
+                <TabsTrigger value="selection">Text Selection</TabsTrigger>
+              </TabsList>
+              <TabsContent value="essentials">
+                <Table>
+                  {essentials.map((shortcut) => (
+                    <Tr>
```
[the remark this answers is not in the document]
pradel: Same comment for all the other `.map` on this file :)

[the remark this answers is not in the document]
gregogun: Ah yes that's right, I forgot the key prop!

PR #399, opened 2022-03-15
feat: header visible on scroll up

gregogun:
This pr adds functionality that allows the user to interact with the header when scrolling up, which allows them to view, save, publish, and access settings when writing long articles.

### DIFF
--- a/sigle/src/modules/editor/EditorHeader.tsx
+++ b/sigle/src/modules/editor/EditorHeader.tsx
@@ -8,6 +8,7 @@ import Tippy from '@tippyjs/react';
 import { Box, Button, Flex, IconButton, Text } from '../../ui';
 import { Story } from '../../types';
 import { useAuth } from '../auth/AuthContext';
+import { useEffect, useState } from 'react';
 
 interface EditorHeaderProps {
   story: Story;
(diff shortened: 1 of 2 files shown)


sigle/src/modules/editor/EditorHeader.tsx
```diff
@@ -26,10 +29,47 @@ export const EditorHeader = ({
...
+
   return (
-    <Flex as="header" justify="between" align="center">
+    <Flex
+      css={{
+        position: 'sticky',
+        transform: scrollDirection === 'down' ? 'translateY(-100%)' : 'none',
+        transition: 'transform 0.5s, padding 0.2s',
+        backgroundColor: '$gray1',
+        top: window.scrollY < 40 ? 'auto' : 0,
```
Akirtovskis, line 63: @gregogun net sure whether this one is responsible for when the header appears on mobile, but when scrolling on mobile it took more scrolling to see the menu than on web.

Akirtovskis (approved):

sigle/src/modules/editor/EditorHeader.tsx
```diff
@@ -26,10 +29,47 @@ export const EditorHeader = ({
   onOpenSettings,
   onSave,
 }: EditorHeaderProps) => {
+  const [scrollDirection, setScrollDirection] = useState<ScrollDirection>(null);
```
pradel, line 32: to avoid one rerender this could be optimised to an object eg:
```
useState({scrollDirection: null, prevOffset: 0})
```

pradel (commented):
One small optimisation that can be done otherwise it's amazing like this 🚀

PR #401, opened 2022-03-17
feat: new dashboard layout

gregogun:
This pr adds the layout for the new dashboard. 

To inspect, navigate to `/dashboard` (will be updated to home route when remaining UI and functionality is added and validated).


sigle/src/modules/layout/components/NewDashboardLayout.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface NavItemProps {
+  href?: string;
+  children: React.ReactNode;
+}
+
+const NavItem = ({ children, ...props }: NavItemProps) => {
+  const router = useRouter();
+  return (
+    <Box
```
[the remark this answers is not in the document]
gregogun: ah yes it is, thanks for spotting that!

sigle/src/modules/layout/components/NewDashboardLayout.tsx
```diff
@@ -0,0 +1,90 @@
...
+      {children}
+    </Box>
+  );
+};
+
+interface NewDashboardLayoutProps {
+  children: React.ReactNode;
+}
+
+export const NewDashboardLayout = ({ children }: NewDashboardLayoutProps) => {
```
pradel, line 61: for this one let's just call it DashboardLayout as the old one is deleted :)

sigle/src/modules/layout/components/NewDashboardLayout.tsx
```diff
@@ -0,0 +1,90 @@
...
+  return (
+    <FullScreen>
+      <AppHeader />
+      <Container
+        css={{
+          display: 'flex',
+          gap: 103,
+          mt: '$10',
+          width: '100%',
+          height: '100%',
```
pradel, line 71: this can probably be removed as there is some content now

pradel (changes requested):
Some visual issues I had while testing:
1. Logo is cropped in the header
Screenshot 2022-03-22 at 17 39 15

2. Wdyt about removing the old buttons, both actions are already in the new header
Screenshot 2022-03-22 at 17 39 27

3. When there is a lot of content, the content is passing on top of the footer 
Screenshot 2022-03-22 at 17 42 28

Screenshot 2022-03-22 at 17 39 42

sigle/src/modules/layout/components/DashboardHeader.tsx
```diff
@@ -131,19 +131,6 @@ export const DashboardPageTitle = ({ title }: DashboardPageTitleProps) => {
...
-            size="large"
-            as="a"
-            href={`/${user?.username}`}
-            target="_blank"
-          >
-            Visit my blog{' '}
-            <EyeOpenIcon width={18} height={18} style={{ marginLeft: 8 }} />
-          </VisitButton>
-          {!loadingCreate && (
-            <Button onClick={handleCreateNewPrivateStory}>New story</Button>
```
pradel, line 144: can we delete the handleCreateNewPrivateStory function in this file?

sigle/src/modules/layout/components/NewDashboardLayout.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export const DashboardLayout = ({ children }: DashboardLayoutProps) => {
+  return (
+    <FullScreen>
+      <AppHeader />
+      <Container
+        css={{
+          display: 'grid',
+          gridTemplateColumns: '1fr 820px 1fr',
```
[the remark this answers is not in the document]
gregogun: Quentin confirmed that width for both is 826px so will update the config :)

sigle/src/modules/layout/components/DashboardLayout.tsx
```diff
@@ -54,11 +60,62 @@ const NavItem = ({ children, ...props }: NavItemProps) => {
   );
 };
 
+interface AccordionNavItemProps {
+  href?: string;
+  children: React.ReactNode;
+}
+
```
gregogun, line 67: created a seperate component from `NavItem` as there were a few differences between the nav items in the sidebar and the ones in the accordion

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -4,14 +4,26 @@ import {
   TwitterLogoIcon,
   DiscordLogoIcon,
   SunIcon,
+  HamburgerMenuIcon,
```
pradel, line 7: regarding the mobile menu of this one, will it be impacted by the coming changes that we will have with the top right corner? It seems to me that these changes should be made in a separate pull request.

pradel (changes requested):
Good job, new layout is 🔥 

- Can you please create another changeset for the tailwind package changes?
- You have some git conflicts on this branch
- The mobile navigation menu should be shown just on the mobile right? 😅 
- Can you check why this warning is happening?
Screenshot 2022-03-30 at 11 41 33

sigle/src/modules/layout/components/DashboardLayout.tsx
gregogun: This approach makes much more sense, and also solves the warning in the console, thanks! 

Following the design, the items still need background colour when matching that particular path. Do you think it's better to add this functionality to them individually or create a wrapping component (similar to above) where that logic can be added? Hope that makes sense 😅

pradel: I think using variants is the way for this :)

sigle/src/modules/layout/components/DashboardLayout.tsx
```diff
@@ -1,257 +1,193 @@
...
+            '@md': {
+              maxWidth: 724,
+            },
+
+            '@lg': {
+              maxWidth: 834,
+            },
+
+            '@xl': {
+              maxWidth: 826,
```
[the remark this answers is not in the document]
gregogun: checked with Quentin and he confirmed that this is intended but should automatically happen with the padding so will remove the hardcoded values :)

sigle/src/modules/layout/components/DashboardLayout.tsx
```diff
@@ -1,257 +1,182 @@
...
+            <NavItem selected={router.pathname === '/settings'}>
+              Settings
+            </NavItem>
+          </Link>
+        </Sidebar>
+        <Box
+          css={{
+            mb: '$5',
+            mx: 'auto',
+            width: '100%',
```
pradel, line 140: These 2 props can probably be removed?

sigle/src/modules/layout/components/DashboardLayout.tsx
```diff
@@ -1,257 +1,182 @@
...
+            collapsible
+            type="single"
+          >
+            <AccordionItem value="item1">
+              <AccordionTrigger>{triggerName}</AccordionTrigger>
+              <AccordionContent>
+                <Link href="/" passHref>
+                  <NavItem hidden={router.pathname === '/'} variant="accordion">
+                    Drafts
+                  </NavItem>
```
pradel, line 154: If the goal is to hide the element would doing smth like this in react make sense?
```
                {router.pathname !== '/' ? <Link href="/" passHref>
                  <NavItem variant="accordion">
                    Drafts
                  </NavItem> : null}
```

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -42,8 +42,18 @@ const Header = styled('header', Container, {
   width: '100%',
 
   '@md': {
+    px: 0,
```
pradel, line 45: Could you please explain the reason of this change?

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -57,7 +67,11 @@ const StatusDot = styled('div', {
   borderRadius: '$round',
 });
 
-export const AppHeader = () => {
+interface AppHeaderProps {
+  children?: React.ReactNode;
```
pradel, line 71: This prop is not used right?

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -114,127 +128,136 @@ export const AppHeader = () => {
...
-              logout
-              <HoverCardArrow />
-            </HoverCardContent>
-          </HoverCard>
-        ) : (
-          <Flex gap="6">
-            <IconButton
-              as="a"
-              href={sigleConfig.twitterUrl}
+    <>
```
pradel, line 131: Looks like fragment only has one element, can you please remove it?

pradel (commented):
Last small nits :)

sigle/src/modules/layout/components/DashboardLayout.tsx
gregogun: Ah yep that makes more sense 😅

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: I think this was from where I was setting the maxWidth at each media query, but didn't take into account the padding that is already there as part of the container. Will remove this and fixed maxWidth sizes 👍

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: Sure, will remove

sigle/src/stitches.config.ts
```diff
@@ -61,6 +61,14 @@ const orangeDark = {
   orange12: '#FF802B',
 };
 
+const red = {
```
pradel, line 64: Are these ones used?

pradel (changes requested):
Last comment and let's merge it tomorrow 🚀

pradel (approved):

PR #406, opened 2022-03-24
fix: code blocks color styles on dark mode

### LINKED ISSUE #405: 🐛 Bug | Issues with code blocks in dark mode
## Description

When you enable dark mode in the app, the code blocks are not working properly.

We need to fix the style of:
- inline code blocks
- multi-line code blocks

Screenshot 2022-03-22 at 15 26 33

gregogun:
Corrects styling for inline and multi-line code blocks on dark mode.

Fixes #405


sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -45,6 +45,10 @@ const fadeInAnimation = keyframes({
 });
 
 const StyledEditorContent = styled(EditorContent, {
+  '& code': {
```
pradel, line 48: as this is related to the rendering of the blocks would be better to move to the tailwind package right?

pradel (approved):

Issue #184, opened 2021-02-26
⚡ Feature | Warning modal when leaving editor with unsaved changes


pradel:
## Description

When the user is on the editor page, show a modal if he click on the logo on the top left saying that some changes are unsaved.

Related to the implementation I found https://github.com/vercel/next.js/issues/2694#issuecomment-732990201 but better to look at other ways too.

Should prevent:
- leaving page using next.js router (eg: clicking arrow on the top left)
- leaving page using browser routing (eg: clicking go back to previous page on chrome)
- leaving page for a new website (eg: change website URL on chrome and press enter)

PR #407, opened 2022-03-28
feat: warning modal for unsaved changes

Akirtovskis:
Closes https://github.com/sigle/sigle/issues/184

Uses the solution proposed within the issue.

Not sure whether the approach used here is perfect, but works pretty smoothly.


pradel:
@gregogun good for review :)

pradel:
@QuentinSaubadu can you please test the behavior to see if it's working as expected?

PR #419, opened 2022-04-07
feat: underline links

### LINKED ISSUE #398: ⚡ Feature | Underline links
## Underline links

As mentioned on call @gregogun last time when we were talking about the link underline on the username register page, let's add underline on links only on hover (not on the editor itself but on the article page) like this:
Capture d’écran 2022-03-14 à 10 59 42

Thanks

gregogun:
This pr updates links on the article page to be underlined when hovered. 

Resolves #398

### DIFF
--- a/sigle/src/modules/publicStory/PublicStory.tsx
+++ b/sigle/src/modules/publicStory/PublicStory.tsx
@@ -80,6 +80,13 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
           "& :where(a strong):not(:where([class~='not-prose'] *))": {
             color: safeSiteColor,
           },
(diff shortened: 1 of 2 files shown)


sigle/src/modules/publicStory/PublicStory.tsx
```diff
@@ -80,6 +80,13 @@ export const PublicStory = ({ story, settings }: PublicStoryProps) => {
           "& :where(a strong):not(:where([class~='not-prose'] *))": {
             color: safeSiteColor,
           },
+          '.ProseMirror': {
```
pradel, line 83: @gregogun should this rather be added in the tailwind package?

PR #427, opened 2022-04-13
fix: underlined links outside of article

### LINKED ISSUE #426: 🐛 Bug | Underline links
Since the push, all the links are underlined but not only in the editor content/frame.

Date/hours, share icons... are now underlined when we hover it.

Capture d’écran 2022-04-13 à 01 06 37

Capture d’écran 2022-04-13 à 01 07 12

## Device

- OS: IOS
- Browser : Chrome
- Version : 99.0.4844.83

gregogun:
This pr fixes the issue of underlined links outside of the article itself (Date, share icons). 

Questions/Concerns: 
- As Quentin mentioned, links are still underlined in the editor so if a user wants to manually underline using `Ctrl/Cmd + U` 2 underlines will appear. Do you have any ideas on how best to approach resolving this @pradel ?

Closes #426

### DIFF
--- a/.changeset/proud-tigers-fail.md
+++ b/.changeset/proud-tigers-fail.md
@@ -2,4 +2,4 @@
 '@sigle/tailwind-style': patch
 ---
 
-Fix underline on links showing up outside of article.
+Fix double underline appearing on editor where a user manually adds an underline to a link.
--- a/.changeset/proud-tigers-fail.md
+++ b/.changeset/proud-tigers-fail.md
@@ -1,0 +1,5 @@
+---
+'@sigle/tailwind-style': patch
+---
+
+Fix underline on links showing up outside of article.
--- a/packages/tailwind-style/dist/tailwind.css
+++ b/packages/tailwind-style/dist/tailwind.css
@@ -846,7 +846,7 @@ Ensure the default browser behavior of the `hidden` attribute.
   color: var(--tw-prose-links);
 }
 
-.prose :where(a:hover):not(:where([class~='not-prose'] *)) {
+.prose :where(.ProseMirror):not(:where([class~='not-prose'] *)) a:hover {
   box-shadow: 0 1px 0 0 currentColor;
 }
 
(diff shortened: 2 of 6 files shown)


pradel:
@gregogun and idea to test to fix the double underline part, looks like underline text is wrapped in the `<u>` HTML tag, so we could try to remove the default text decoration and apply the same boxShadow as hover links instead?

packages/tailwind-style/dist/tailwind.css
```diff
@@ -846,7 +846,7 @@ Ensure the default browser behavior of the `hidden` attribute.
   color: var(--tw-prose-links);
 }
 
-.prose :where(a:hover):not(:where([class~='not-prose'] *)) {
+.prose :where(.ProseMirror):not(:where([class~='not-prose'] *)) a:hover {
```
pradel, line 849: This tailwind package should be framework agnostic so users not using prosemirror to render the text should not be impacted by any changes made to this package.
I think a better approach is to use the class `className="not-prose"` to remove the styling on the affected blocks.

gregogun: for some reason, adding `className="not-prose"` doesn't seem to remove styling on affected blocks. Looking at the devtools, the text decoration is still appearing so wondering if there is another way to remove this?

pradel: It definitely should :( could you please push the not working part so I can have a look?

pradel:
@gregogun just pushed a commit, looks like it needs to be applied to one of the parents containing the link in order to work :)

.changeset/proud-tigers-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sigle/tailwind-style': patch
+---
+
+Fix underline on links showing up outside of article.
```
pradel, line 5: could you update the message of this one and add a changeset for the app please?

pradel (approved):
🚀

PR #428, opened 2022-04-14
feat: add clearer interactive states for dropdown

gregogun:
Updates hover and active states to make interactions between dropdown items clearer for users with low-vision.

### DIFF
--- a/sigle/src/stitches.config.ts
+++ b/sigle/src/stitches.config.ts
@@ -69,10 +69,14 @@ const greenDark = {
 };
 
 const red = {
+  red4: '#FFEAEA',
+  red5: '#FFDBDB',
   red11: '#F43939',
 };
 
 const redDark = {
+  red4: '#4C0808',
+  red5: '#570909',
   red11: '#FF6C6C',
 };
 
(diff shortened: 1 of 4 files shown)


sigle/src/stitches.config.ts
```diff
@@ -69,10 +69,14 @@ const greenDark = {
 };
 
 const red = {
```
gregogun, line 71: I've just added the colors being used in this pr to the ds - just wondering if this is fine or if it's better practice to add all of the green and red colors as they are now in the ds on figma?

sigle/src/ui/DropdownMenu.ts
```diff
@@ -5,6 +5,7 @@ export const DropdownMenu = DropdownMenuPrimitive.Root;
 export const DropdownMenuTrigger = DropdownMenuPrimitive.Trigger;
 export const DropdownMenuSeparator = styled(DropdownMenuPrimitive.Separator, {
   height: 1,
+  mx: '-$3',
```
gregogun, line 8: make separator full width despite parent padding

sigle/src/stitches.config.ts
pradel: Should be good like this 👍

pradel (approved):
Nice 👌

PR #431, opened 2022-04-14
feat: add dropdown header

gregogun:
Replaces hover card to now display a dropdown menu when clicking on username.

### DIFF
--- a/sigle/src/modules/layout/components/AppHeader.tsx
+++ b/sigle/src/modules/layout/components/AppHeader.tsx
@@ -165,16 +165,15 @@ export const AppHeader = () => {
                 tabIndex={0}
                 css={{
                   cursor: 'pointer',
-                  p: '$2',
+                  p: '$3',
                   br: '$1',
                   '&:hover': { backgroundColor: '$gray4' },
-                  '&:active': { backgroundColor: '$gray5' },
                 }}
-                gap="1"
+                gap="2"
                 align="center"
               >
                 <StatusDot />
-                <Text>{user.username}</Text>
+                <Text size="sm">{user.username}</Text>
               </Flex>
             </DropdownMenuTrigger>
             <DropdownMenuContent sideOffset={8}>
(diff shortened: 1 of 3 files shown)


sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -187,12 +193,29 @@ export const AppHeader = () => {
...
+                  <a>Dashboard</a>
+                </Link>
+              </DropdownMenuItem>
+              <DropdownMenuItem>
+                <Link href="/settings" passHref>
+                  <a>Settings</a>
+                </Link>
+              </DropdownMenuItem>
+              <DropdownMenuSeparator />
+              <DropdownMenuItem color="red" onClick={handleLogout}>
```
pradel, line 214: @gregogun I think the theme switch under the flag is missing :)

gregogun: ah I thought this was to be added after the dark mode is fully integrated but then remembered we have the experimental flag! Will add it now :) 

Should I remove the existing one theme toggle button, or should it be left? 
Screenshot 2022-04-19 at 08 27 35

pradel: You can remove it 👍

pradel:
Couple of questions for both of you @QuentinSaubadu @gregogun 
1. Should there be a hover on the dropdown menu trigger? Like on the left menu?
2. Should clicking on my blog open the link in a new tab as it's done at the moment?

QuentinSaubadu:
> 

1- The trigger definitely should have a hover, like other incoming links in the header like "Discover". I'm working on it.

2- Yes, clicking on "my blog" will open a new tab. Then when we move to the profile page it will just redirect to the profile page on the same tab.

3- The "visit my blog" is actually removed from the new header (see header components)

pradel:
Last question I didn't see before but this is the logged in menu, should we leave the theme toggle on the right side as it is now for non logged in users only?

gregogun:
Yeah that makes sense 👍

pradel (commented):
@gregogun found a small issue, clicking on the right side of the menu item is not redirecting to the page. Maybe the item menu should be `as="a"` like this the padding is working for the link too?
Screenshot 2022-04-20 at 14 31 28

gregogun:
ahh yes I can see the issue there, will update 👍

pradel (approved):
Good job 🚀

PR #434, opened 2022-04-21
feat: redesign story cards

### LINKED ISSUE #412: ⚡ Feature | Story cards update
We need to change the story card design according to this Figma file: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/?node-id=1038%3A1646

- We need to remove the top line on the 1st card of the column. 
- The description is removed on the mobile version
- The "more" button dropdown is available here: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/?node-id=1841%3A1794
Capture d’écran 2022-04-01 à 15 23 01
- Draft label is a new component to be integrated on the draft cards (component name: Draft Dashboard Card): https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/?node-id=1116%3A2067
Capture d’écran 2022-04-01 à 15 26 42
- Title and description are 2 lines long maximum. If the title or the description is longer, add "..." at the end.
Capture d’écran 2022-04-01 à 15 27 27
- The card specs can be changed on Figma like this: 

https://user-images.githubusercontent.com/7244762/161273457-6f47e9e3-83a8-42dd-b00e-9717fd6054b3.mov

gregogun:
This pr updates the story cards to the new redesign. 

Closes #412

### DIFF
--- a/sigle/src/modules/home/components/StoryItem.tsx
+++ b/sigle/src/modules/home/components/StoryItem.tsx
@@ -1,7 +1,5 @@
 import React from 'react';
 import Link from 'next/link';
-import styledC from 'styled-components';
-import tw from 'twin.macro';
 import { StarFilledIcon } from '@radix-ui/react-icons';
 import { DotsHorizontalIcon } from '@radix-ui/react-icons';
 import format from 'date-fns/format';
--- a/sigle/src/modules/home/components/StoryItem.tsx
+++ b/sigle/src/modules/home/components/StoryItem.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import Link from 'next/link';
-import styled from 'styled-components';
+import styledC from 'styled-components';
 import tw from 'twin.macro';
 import { StarFilledIcon } from '@radix-ui/react-icons';
 import { DotsHorizontalIcon } from '@radix-ui/react-icons';
@@ -20,37 +20,47 @@ import {
   Heading,
   Text,
   DropdownMenuSeparator,
+  IconButton,
+  Box,
 } from '../../../ui';
 import { SubsetStory, BlockstackUser } from '../../../types';
+import { styled } from '../../../stitches.config';
 
-const StoryContainer = styled.div`
-  ${tw`py-4 lg:py-8 border-b border-solid border-grey lg:flex`};
-`;
+const StoryContainer = styled('div', {
+  display: 'flex',
+  borderBottom: '1px solid $colors$gray6',
+  py: '$7',
+  gap: '$5',
 
-const StoryTitleContainer = styled.div`
-  ${tw`flex`};
-`;
+  '& img': {
+    transform: 'none',
+    transition: 'transform .5s .1s',
+  },
 
-const StarIcon = styled(StarFilledIcon)`
-  ${tw`ml-2`};
-  color: rgb(255, 199, 0);
-`;
+  '&:hover': {
+    '& img': {
+      transform: 'scale(1.1)',
+      transition: 'transform .5s .1s',
+    },
+    '& h4': {
+      color: '$gray12',
+    },
+  },
+});
 
-const StoryTitle = styled.h3`
-  ${tw`flex text-xl no-underline text-black mr-3`};
-`;
+const StoryImage = styled('img', {
+  objectFit: 'cover',
+  objectPosition: 'center',
+  width: 80,
+  height: 58,
+  zIndex: -1,
+  position: 'relative',
 
-const StoryDate = styled.p`
-  ${tw`mt-1 text-sm italic text-grey-dark`};
-`;
-
-const StoryText = styled.p`
-  ${tw`mt-4 text-grey-dark`};
-`;
-
-const StoryImage = styled.div`
-  ${tw`hidden xl:block bg-no-repeat bg-cover bg-center h-32 w-56 ml-8`};
-`;
+  '@md': {
+    width: 180,
+    height: 130,
+  },
+});
 
 interface Props {
   user: BlockstackUser;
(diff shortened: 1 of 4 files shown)


sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -20,37 +20,47 @@ import {
...
+const StoryContainer = styled('div', {
+  display: 'flex',
+  borderBottom: '1px solid $colors$gray6',
+  py: '$7',
+  gap: '$5',
 
-const StoryTitleContainer = styled.div`
-  ${tw`flex`};
-`;
+  '& img': {
```
gregogun, line 33: enable image zoom & brighten story title when hovering anywhere on card

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -97,81 +107,223 @@ export const StoryItem = ({
...
   return (
     <React.Fragment>
-      <StoryContainer>
-        <div style={{ flex: 1 }}>
-          <StoryTitleContainer>
-            <StoryTitle>
+      <StoryContainer
+        css={{
+          '@md': {
+            gap: !story.coverImage && story.featured ? '$5' : '$7',
```
gregogun, line 111: reduce the gap between `featured` icon and text content if there is no image

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -97,81 +107,223 @@ export const StoryItem = ({
...
+            >
+              <Box
+                as="span"
+                css={{
+                  display: 'inline-block',
+                  position: 'relative',
+                  overflow: 'hidden',
+                  borderRadius: '$1',
+
+                  '&::before': {
```
gregogun, line 146: enables overlay on image hover

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -97,81 +105,222 @@ export const StoryItem = ({
...
+                  <Heading
+                    css={{
+                      fontSize: '$2',
+                      lineHeight: '20.4px',
+                      alignSelf: 'center',
+                      fontWeight: 600,
+                      overflow: 'hidden',
+                      display: '-webkit-box',
+                      '-webkit-line-clamp': 3,
+                      '-webkit-box-orient': 'vertical',
```
pradel, line 209: Looks like this one is not supported by firefox https://caniuse.com/?search=-webkit-box-orient, is it really used?

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -97,81 +105,222 @@ export const StoryItem = ({
...
+                <a>
+                  <Heading
+                    css={{
+                      fontSize: '$2',
+                      lineHeight: '20.4px',
+                      alignSelf: 'center',
+                      fontWeight: 600,
+                      overflow: 'hidden',
+                      display: '-webkit-box',
+                      '-webkit-line-clamp': 3,
```
pradel, line 208:
```suggestion
                      'line-clamp': 3,
```
We should remove the webkit prefix as some browsers might not support it :)

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -97,81 +105,222 @@ export const StoryItem = ({
...
             <Link
               href="/stories/[storyId]"
               as={`/stories/${story.id}`}
               passHref
             >
-              <a>{format(story.createdAt, 'HH:mm dd MMMM yyyy')}</a>
+              <Text size="xs" as="a">
+                {format(story.createdAt, 'MMMM dd, yyyy ')}
+                at
+                {format(story.createdAt, ' h:mmaaa')}
```
pradel, line 293: using `h` is not prefixed with 0 for numbers < 10, but using `mm` does prefix it. To be consistent let's decide on prefixing both or not prefixing both :)

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -20,37 +18,47 @@ import {
...
+  },
 
-const StarIcon = styled(StarFilledIcon)`
-  ${tw`ml-2`};
-  color: rgb(255, 199, 0);
-`;
+  '&:hover': {
+    '& img': {
+      transform: 'scale(1.1)',
+      transition: 'transform .5s .1s',
```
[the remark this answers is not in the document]
gregogun: ah sorry this one is actually what is enabling image zoom when hovering on card - `'& img'` styles declared above are ensuring zoom out is matching the speed of the zoom in

sigle/src/modules/home/components/StoryItem.tsx
```diff
@@ -20,37 +18,47 @@ import {
...
+    },
+    '& h4': {
```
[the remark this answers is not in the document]
gregogun: this is changing the story title but I think the effect is very subtle, could discuss with Quentin if we need it?

sigle/src/modules/home/components/StoryItem.tsx
gregogun: just tested and when I remove it, it removes the clamp effect. Do you know if it's possible to have both?

sigle/src/modules/home/components/StoryItem.tsx
gregogun: just tested and is needed for clamping effect. After some reading, I think it can be enabled in firefox by using the `moz` prefix?

sigle/src/modules/home/components/StoryItem.tsx
gregogun: ah I see, so would it make sense to do something like `h:maaa` for not prefixed or  `hh:mmaaa` for prefixed ?

pradel: exactly 👍

sigle/src/modules/home/components/StoryItem.tsx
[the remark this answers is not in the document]
pradel: If it's on purpose no problem for me I was just not able to spot the difference :p

sigle/src/modules/home/components/StoryItem.tsx
gregogun: hmm not sure but will have a look into it :)

gregogun:
found from [this](https://css-tricks.com/almanac/properties/l/line-clamp/) article that browser support is now generally good for webkit prefix; did some testing on firefox and clamping functionality is working well :)

PR #447, opened 2022-04-27
fix: dialog portal issue

gregogun:
Although the previous fix took care of the header appearing over settings at the top of the page, once further down the page and trying to access the settings, the bug remained like so:
Screenshot 2022-04-27 at 12 24 54

After some digging, I came across [this](https://discord.com/channels/752614004387610674/803656530259738674/918963737610838078) from the Radix help discord channel - so following the advice given there to set a new stacking context at root in conjunction with the Dialog Portal has now solved the issue :)

### DIFF
--- a/.changeset/sixty-dragons-attack.md
+++ b/.changeset/sixty-dragons-attack.md
@@ -1,0 +1,5 @@
+---
+'@sigle/app': patch
+---
+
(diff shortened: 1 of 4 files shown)


pradel (approved):
Good find!

PR #461, opened 2022-05-05
feat: analytics stories frame

gregogun:


sigle/src/modules/analytics/PublishedStoriesFrame.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const firstPageIndex = (currentPage - 1) * itemSize;
+    const lastPageIndex = firstPageIndex + itemSize;
+    setCurrentStories(stories.slice(firstPageIndex, lastPageIndex));
+  }, [currentPage]);
+
+  return (
+    <>
+      {currentStories && (
+        <>
+          <Box css={{ mb: '$3', height: 476 }}>
```
gregogun, line 32: fixed height to prevent pagination component shifting if the next or previous page has more or less items

sigle/src/modules/analytics/PublishedStoryItem.tsx
```diff
@@ -0,0 +1,116 @@
...
+      className="story-item"
+      align="center"
+      css={{
+        borderTop: '1px solid $colors$gray6',
+        height: 68,
+        cursor: 'pointer',
+        position: 'relative',
+        opacity: 0,
+
+        '& svg': {
```
gregogun, line 49: arrow transitions on hover

sigle/src/modules/analytics/PublishedStoryItem.tsx
```diff
@@ -0,0 +1,116 @@
...
+          '& div:first-child': {
+            display: 'block',
+            backgroundColor: '$gray11',
+          },
+        },
+
+        '& div:first-child': {
+          position: 'absolute',
+          width: '$1',
+          top: -1,
```
gregogun, line 68: styles here are ensuring indicator is positioned over top border and top border of element below:

Screenshot 2022-05-05 at 12 40 36

sigle/src/modules/analytics/PublishedStoryItem.tsx
```diff
@@ -0,0 +1,116 @@
...
+    play();
+  }, []);
+
+  return (
+    <Flex
+      className="story-item"
+      align="center"
+      css={{
+        borderTop: '1px solid $colors$gray6',
+        height: 68,
```
gregogun, line 44: make height consistent for cards without an image

pradel:
@gregogun can you merge main into the branch to fix the CI ? :)

sigle/src/modules/analytics/Pagination.tsx
```diff
@@ -0,0 +1,55 @@
+import { ChevronLeftIcon, ChevronRightIcon } from '@radix-ui/react-icons';
+import { SubsetStory } from '../../types';
+import { Flex, IconButton } from '../../ui';
+
+interface PaginationProps {
+  currentPage: number;
+  onPageChange: (page: number) => void;
+  stories: SubsetStory[];
+  currentStoryItems: SubsetStory[];
```
pradel, line 9: To make is more like a generic component that can later be reused, you could remove these props and just add a boolean `hasNextPage` for example, that way you can reuse it for referrers and other :)

sigle/src/modules/analytics/components/Analytics.tsx
```diff
@@ -0,0 +1,22 @@
...
+  stories: SubsetStory[] | null;
+  loading: boolean;
+}
+
+export const Analytics = ({ stories, loading }: AnalyticsProps) => {
+  const nbStoriesLabel = loading ? '...' : stories ? stories.length : 0;
+
+  return (
+    <DashboardLayout>
+      <Text as="h3" size="sm" css={{ mb: '$3', fontWeight: 600 }}>
```
pradel, line 16: If it's H3 maybe we should rather use the Heading component ?

sigle/src/modules/analytics/PublishedStoriesFrame.tsx
```diff
@@ -0,0 +1,47 @@
...
+}: PublishedStoriesFrameProps) => {
+  const [currentStories, setCurrentStories] = useState<SubsetStory[] | null>(
+    stories
+  );
+  const [currentPage, setCurrentPage] = useState<number>(1);
+
+  // how many stories we should show on each page
+  let itemSize = 7;
+
+  useEffect(() => {
```
pradel, line 22: Regarding the `useMemo` question, I don't think we need that optimisation for performance reasons with big lists, but using useMemo here means that we can avoid a setState and avoid a rerender so using it for this sounds good :)

pradel (changes requested):
Great one @gregogun 😍 , some super small comments but other than this it's perfect 👌

sigle/src/modules/analytics/PublishedStoriesFrame.tsx
gregogun: oh right I see, thanks for clarifying - will update 👍

PR #467, opened 2022-05-10
feat: analytics referrers

### LINKED ISSUE #451: ⚡ Feature | Create referrers for Analytics
Step 2 for the analytic page is the referrer frame.

we have to show the author from which sites his readers are coming and the number of views for each site:
Capture d’écran 2022-05-02 à 12 04 01

the grey bar represents the number of views in terms of percentage compared to the other sites in the list.
The maximum (100%) corresponds to the total width of the block, while the minimum is calculated according to the total number of referrers displayed.

Example: 
Capture d’écran 2022-05-02 à 12 28 34

L&R SWITCH
The L&R switch should have the same transition as the story card frame #450

gregogun:
Adds referrers with dummy data to the analytics page. 

Enable analytics page by adding `?experimentalAnalyticsPage=true` at the end of the root URL.

Closes #451

### DIFF
--- a/.changeset/selfish-bananas-lie.md
+++ b/.changeset/selfish-bananas-lie.md
@@ -1,5 +1,5 @@
 ---
-'@sigle/app': minor
+'@sigle/app': patch
 ---
 
-Create UI with dummy data for referrers on analytics page.
+(Beta) Create UI with dummy data for referrers on analytics page.
--- a/.changeset/selfish-bananas-lie.md
+++ b/.changeset/selfish-bananas-lie.md
@@ -1,0 +1,5 @@
+---
+'@sigle/app': minor
+---
+
+Create UI with dummy data for referrers on analytics page.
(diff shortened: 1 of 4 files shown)


sigle/src/modules/layout/components/DashboardLayout.tsx
```diff
@@ -12,6 +12,40 @@ import {
...
+
+  variants: {
+    layout: {
+      default: {
+        '@xl': {
+          gridTemplateColumns: '1fr 826px 1fr',
+        },
+      },
+      wide: {
+        gridTemplateColumns: '1fr 5fr',
```
gregogun, line 40: working solution to get analytics content to fill the rest of the available space, whilst keeping the same relative distance to the sidebar as on other pages. Just wondering if there may be a better way to achieve this?

pradel: Looks like it's working fine that way 👍

.changeset/selfish-bananas-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sigle/app': minor
+---
+
+Create UI with dummy data for referrers on analytics page.
```
pradel, line 5: For changes behind a flag that users can't use directly can we start the message and prefix with (Beta) (so they are sorted in the changelog) and always mark them as patch? Are you fine with this idea?

gregogun: Okay yeah that sounds good - will update it :)

PR #475, opened 2022-05-16
feat: chart analytics

### LINKED ISSUE #452: ⚡ Feature | Chart design for Analytics page
Step 3 for the Analytics page is to display the graph with 2 data, total visitors and total views.

- The total number of visitors is the total number of unique users who visited a given profile.

- The total number of views corresponds to all the user clicks on a given profile (each time a user clicks on a published story for example).

On this issue, we are focusing on the chart itself and **not on the number display**:
Capture d’écran 2022-05-02 à 12 41 52

**Use that CodeSandBox to make the chart:**
https://codesandbox.io/s/hrje5

Link to Figma page: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=1787%3A12400

Estimate: 3D

gregogun:
Adds tabbable graph views to display the views and visits stats for the user's blog. 

Closes #452


sigle/src/modules/analytics/stats/StatsFrame.tsx
```diff
@@ -0,0 +1,166 @@
...
+    });
+
+    setData(stats);
+  };
+
+  const fetchMonthlyStats = async () => {
+    setData(initialRange);
+
+    const statsRes = await fetch(
+      `${baseUrl}/api/analytics/historical?dateFrom=${format(
```
pradel, line 69: I think this should work fine too :) 
```suggestion
      `/api/analytics/historical?dateFrom=${format(
```

sigle/src/modules/analytics/stats/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { defaultStyles } from '@visx/tooltip';
```
pradel, line 1: as the utils are just for the tooltip should we rename it to tooltip.ts? or tooltipUtils?

sigle/src/modules/analytics/stats/StatsFrame.tsx
```diff
@@ -0,0 +1,178 @@
...
+          <TabsContent value="weekly">
+            <Box
+              css={{
+                mb: '$5',
+                position: 'relative',
+                width: '100%',
+                height: 400,
+              }}
+            >
+              <StatsChart type="week" data={data} />
```
pradel, line 147: As we are using tabs, the StatsChart element is recreated and destroyed every time the type is changing. I am wondering if we were rendering the same view but just changing the props passed down to StatsChart, would the graph change the values with a nice animation?

sigle/src/modules/analytics/stats/StatsFrame.tsx
```diff
@@ -0,0 +1,178 @@
...
+        fetchWeeklyStats();
+        break;
+      case 'monthly':
+        fetchMonthlyStats();
+        break;
+      case 'all':
+        fetchAllTimeStats();
+        break;
+      default:
+        fetchWeeklyStats();
```
pradel, line 117: Default should probably throw an error as this should not happen :)

sigle/src/modules/analytics/stats/StatsFrame.tsx
```diff
@@ -0,0 +1,178 @@
...
+export const StatsFrame = () => {
+  const [data, setData] = useState<StatsData[]>(initialRange);
+
+  const baseUrl = window.location.origin;
+
+  useEffect(() => {
+    fetchWeeklyStats();
+  }, []);
+
+  const fetchWeeklyStats = async () => {
```
pradel, line 43: by looking at the code it feels like fetchWeeklyStats, fetchMonthlyStats and fetchAllTimeStats could be merged into one function as they are doing the same thing, wdyt?

sigle/src/types/index.ts
```diff
@@ -103,3 +103,26 @@ export interface SettingsFile {
    */
   siteTwitterHandle?: string;
 }
+
+export interface AnalyticsHistoricalData {
```
pradel, line 107: As these types are only used by the analytics, the types should probably stay within the stats module

sigle/src/types/index.ts
```diff
@@ -103,3 +103,26 @@ export interface SettingsFile {
...
+  pageviews: number;
+}
+
+export type AnalyticsHistoricalResponse = {
+  historical: AnalyticsHistoricalData[];
+  stories: AnalyticsStoriesData[];
+};
+
+export interface StatsData {
+  value: number;
```
pradel, line 125: to avoid confusion, should this be renamed to pageviews?

pradel (changes requested):
Good job this one was a big one 🚀 

I found the following bug while testing, my mouse is on the line, showing the 83 views tooltip, but the 83 pick is actually on the right side. So the tooltip is not showing exactly what you are hovering.

Screenshot 2022-05-20 at 10 12 16

sigle/src/modules/analytics/stats/StatsFrame.tsx
gregogun: agreed, will merge them :)

sigle/src/types/index.ts
gregogun: would it be worth putting re-used types like `StatsData` in a separate types file within the stats module?

pradel: Sure 👍

sigle/src/modules/analytics/stats/StatsFrame.tsx
gregogun: Great idea - updated and this is working much better (no longer a flash of no content between tabs), thanks!

pradel (approved):
Will merge on monday to avoid a Friday deployment :D

Issue #491, opened 2022-05-23
⚡ Feature | Analytics error


pradel:
## Description

When the API returns an error we should display it so the users know something wrong happened.

Idea: Show some red alert message on top of the view.

QuentinSaubadu:
Done: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/?node-id=1774%3A12059
<img width="1041" alt="Capture d’écran 2022-05-24 à 17 22 53" src="https://user-images.githubusercontent.com/7244762/170072940-30fabed5-3787-45ca-8949-a8816fb82aa0.png">

The frame is full width but can work with multi-lines.
Text: 11/Subheading
Background: 5

Issue #456, opened 2022-05-02
⚡ Feature | Display the referrers in a story's page on Analytics


QuentinSaubadu:
Use the code from the issue #451 to display the referrers of a given story.
<img width="761" alt="Capture d’écran 2022-05-02 à 18 38 28" src="https://user-images.githubusercontent.com/7244762/166285742-686f1f83-64ff-4109-843a-27453d3edb3b.png">

Estimate: 1D

Issue #455, opened 2022-05-02
⚡ Feature | Display the graph in a story's page on Analytics


QuentinSaubadu:
Use the Graph you created on issue #452 to display the total view of a given story.

<img width="699" alt="Capture d’écran 2022-05-02 à 18 35 30" src="https://user-images.githubusercontent.com/7244762/166281894-7d5a86f8-f41d-4273-9aa4-7749a0a1e362.png">

Estimate: 1D

PR #494, opened 2022-05-24
feat: individual story analytics

### LINKED ISSUE #454: ⚡ Feature | Create the open version of the story card
Clicking on a Story card (Step 1 #450) sends the user to another page (https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=1856%3A12287).

This page is here to give you more detailed information about a particular story.
To allow the user to have an eye on the story he is viewing, we re-display the card (#450) at the top of the page but slightly modified.
Capture d’écran 2022-05-02 à 13 03 09

The number of views is removed and the arrow goes from right to left side.

The rest remains the same.

ARROW ANIMATION
As you did on the arrow animation on issue #450, the arrow remains animated in hover but in reverse, from left to right in hover:

https://user-images.githubusercontent.com/7244762/166224671-dbf5f1c4-ae33-403b-8417-88e546a811f2.mov

The whole card is clickable in order to get back to the main analytics page.

gregogun:
Adds an analytics page for each story displaying views, visitors, and referrers. 
Replaces placeholder data with real API data, for the story views and referrers.

Closes #454 #455 #456 #491


sigle/src/modules/analytics/stats/StatsTotal.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const numberWithCommas = (x: string): string => {
+  return new Intl.NumberFormat('en-US').format(Number(x)).toString();
+};
+
+export const StatsTotal = ({ data }: StatsTotalProps) => {
+  const [totalViewsAndVisitors, setTotalViewsAndVisitors] =
+    useState<TotalViewsAndVisitorsProps>();
+
+  useEffect(() => {
```
pradel, line 22: I feel like this could be changed to `useMemo` so we can avoid rendering it twice :)

sigle/src/modules/analytics/stats/StatsFrame.tsx
```diff
@@ -90,49 +62,32 @@ export const StatsFrame = () => {
...
-          visits: previousValue.visits + currentValue.visits,
-        };
-      },
-      { pageviews: 0, visits: 0 }
-    );
-
-    setData(stats);
-    setStatType(value);
-    setTotalViewsAndVisitors(viewTotal);
+  const checkStatType = (value: string) => {
```
pradel, line 68: can we type value so it's not just a string with ts please ?

sigle/src/modules/analytics/stats/StatsFrame.tsx
```diff
@@ -144,16 +99,15 @@ export const StatsFrame = () => {
...
             css={{
-              mb: '$5',
+              mb: '$8',
               position: 'relative',
               width: '100%',
               height: 400,
             }}
           >
-            <StatsChart type={statType} data={data} />
+            <StatsChart type={statType} data={data!} />
```
pradel, line 110: let's try to not use `!` here, if there is no data then the graph should render empty correct? Maybe better to say that StatsChart data prop is not required?

sigle/src/modules/analytics/stats/StatsError.tsx
```diff
@@ -0,0 +1,28 @@
+import { CrossCircledIcon } from '@radix-ui/react-icons';
```
pradel, line 1: Just a note but we should see with Quentin to make the alerts part of the design-system so we can reuse them easily on other views if needed. Not to do now, but smth to think about for the future.

sigle/src/modules/analytics/containers/StoryItemAnalytics.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+interface StorItemAnalyticsProps {
+  storyId: string;
+}
+export const StoryItemAnalytics = ({ storyId }: StorItemAnalyticsProps) => {
+  const [stories, setStories] = useState<SubsetStory[] | null>(null);
+
+  const loadStoryFile = async () => {
+    try {
+      const file = await getStoriesFile();
```
pradel, line 15: should we rather load the individual story file? Also let's please use useQuery there :)

sigle/src/modules/analytics/components/StoryItemAnalytics.tsx
```diff
@@ -0,0 +1,150 @@
...
+        break;
+      case 'all':
+        url = allTimeStatsUrl;
+        break;
+
+      default:
+        throw new Error('No value received.');
+    }
+
+    const statsRes = await fetch(url);
```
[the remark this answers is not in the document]
pradel: Yeah for fetch you have to check the status code is 200 or throw an error :)

sigle/src/modules/analytics/ReferrersFrame.tsx
```diff
@@ -1,80 +1,26 @@
...
-  },
-];
+  const statsRes = await fetch(url);
+  const referrerData: ReferrersResponse = await statsRes.json();
+  return referrerData;
+};
 
 export const ReferrersFrame = () => {
   const [currentPage, setCurrentPage] = useState<number>(1);
+  const { data: referrers } = useQuery<ReferrersResponse, Error>(
```
pradel, line 20: let's also handle the error here :)

sigle/src/modules/analytics/PublishedStoryItem.tsx
```diff
@@ -1,25 +1,52 @@
...
+  individualStory?: boolean;
 }
 
-export const PublishedStoryItem = ({ story }: PublishedStoryItemProps) => {
+export const PublishedStoryItem = ({
+  story,
+  onClick,
+  individualStory = false,
+}: PublishedStoryItemProps) => {
+  const { data } = useQuery<number, Error>(['fetchStoryViews'], () =>
```
pradel, line 47: here you are calling the same endpoint as the graph right? To avoid asking the server to do the computation twice, would using the same cache key with useQuery make only one call to the backend?

pradel (changes requested):
Good job!

Some issues I found while testing :)

On the graph:
- when number total of visitors is zero, "--" is shown while I think we should show "0"

On the referrer component:
- for longer referrers, the title is split in 2 lines but there is enough space for just one. If the title is longer than the content it should end with "..." but stay on just one line.
Screenshot 2022-05-27 at 10 53 35


On the individual story stats page:
- missing the title for the referrer section

sigle/src/modules/analytics/PublishedStoryItem.tsx
gregogun: checked the docs and yeah that should work thx :)

sigle/src/modules/analytics/stats/StatsFrame.tsx
gregogun: with this one, because there's the initial range already added as a placeholder, my assumption was that there should be always be data - is it because the data from the API could be returned undefined? If so I'm wondering the best place to have the conditional logic, as at this level it also removes the axis

pradel: The API should not return undefined (unless there is a fetch error ofc), `data` is undefined on the first render in any case right?

sigle/src/modules/analytics/stats/StatsError.tsx
pradel: Do you mind creating an issue so we don't forget about this one please? :)

sigle/src/modules/analytics/ReferrersFrame.tsx
pradel: You forgot this one ;)

sigle/src/config.ts
```diff
@@ -2,6 +2,10 @@ export const sigleConfig = {
   env: process.env.NODE_ENV,
   isServer: typeof window === 'undefined',
   appUrl: process.env.APP_URL as string,
+  baseUrl:
```
pradel, line 5: let's inject the values from the .env files :)

gregogun: hmm seems like the value is coming up as undefined when I try to pull them from .env.local - do you know what could be potentially causing them to show up as undefined?

pradel: what's the name of the variable?

pradel: You might want to have a look at sigle/next.config.js to inject them :)

gregogun: yep that did it, thanks :)

sigle/next.config.js
```diff
@@ -13,6 +13,8 @@ const nextConfig = {
     FATHOM_SITE_ID: process.env.FATHOM_SITE_ID,
     NEXT_PUBLIC_SENTRY_DSN: process.env.NEXT_PUBLIC_SENTRY_DSN,
     NEXT_PUBLIC_POSTHOG_TOKEN: process.env.NEXT_PUBLIC_POSTHOG_TOKEN,
+    API_URL_DEV: process.env.API_URL_DEV,
```
pradel, line 16: Why do we have 2 env variables here? Wdyt about merging both of them to `API_URL`?

sigle/src/modules/analytics/PublishedStoryItem.tsx
```diff
@@ -1,25 +1,53 @@
...
-  ml: '$2',
 });
 
+const baseUrl = sigleConfig.baseUrl;
+
+const fetchStoryViews = async (storyId: string) => {
+  const url = `${baseUrl}/api/analytics/historical?dateFrom=${FATHOM_MAX_FROM_DATE}&dateGrouping=month&storyId=${storyId}`;
+
+  const statsRes = await fetch(url);
+  const statsData: AnalyticsHistoricalResponse = await statsRes.json();
```
pradel, line 27: could you please throw an error in case smth is wrong with fetch? :)

pradel (changes requested):

sigle/next.config.js
pradel: Why do you want to have a check there? As the config is coming from the .env file (which is different in dev and prod) the value will not be the same.
We can just declare one value, the same way as we do with `APP_URL` :)

gregogun: ahh right yes of course, totally makes sense :)

PR #496, opened 2022-05-24
feat: create sigle API server

pradel:
We will be using fastify for the API server.
For the hosting we will be using render.com


render.yaml
```diff
@@ -0,0 +1,20 @@
+services:
+  # Sigle api server instance
+  - name: sigle-api
+    type: web
+    env: node
+    region: oregon
+    plan: Starter
+    # TODO change this before merging
```
pradel, line 8: this needs to be changed

PR #538, opened 2022-06-06
fix: fix header width on profile page

### LINKED ISSUE #517: 🐛 Bug | Fix header's width on profile page
We need the header to be the same width everywhere.

On the profile page, the header is in a container of 960px where it should be 1280px.

See an example of bugged header : https://app.sigle.io/jackbinswitch.btc
Should be like this : https://app.sigle.io/jackbinswitch.btc/3IlTtz4ymc7PICXjCCQw0

gregogun:
This pr does the following:
- Fixes incorrect header width on profile page
- Refactors deprecated `Text` components to use new `Typography` components
- Replaces all styled components on the page with stitches

Closes #517

### DIFF
--- a/sigle/src/modules/publicHome/components/PublicHome.tsx
+++ b/sigle/src/modules/publicHome/components/PublicHome.tsx
@@ -2,11 +2,10 @@ import React from 'react';
 import { useRouter } from 'next/router';
 import { NextSeo } from 'next-seo';
 import { StoryFile, SettingsFile } from '../../../types';
-import { Container } from '../../../components';
 import { PublicStoryItem } from './PublicStoryItem';
 import { PoweredBy } from '../../publicStory/PoweredBy';
 import { AppHeader } from '../../layout/components/AppHeader';
-import { Box, Flex, Typography } from '../../../ui';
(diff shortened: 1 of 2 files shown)


sigle/src/modules/publicHome/components/PublicHome.tsx
```diff
@@ -116,56 +106,61 @@ export const PublicHome = ({ file, settings }: PublicHomeProps) => {
...
         }}
       />
-
-      <Container>
-        <AppHeader />
-        <Header>
-          {settings.siteLogo && (
-            <HeaderLogo src={settings.siteLogo} alt={`${siteName} logo`} />
+      <AppHeader />
+      <Header>
```
pradel, line 110: I think the header should be inside a container, big images won't overflow that way and on mobile you have some padding around :)

pradel (changes requested):

sigle/src/modules/publicHome/components/PublicHome.tsx
gregogun: ah yes that makes sense - will update :)

PR #541, opened 2022-06-07
feat: compare plans page

gregogun:


sigle/src/modules/settings/plans/ComparePlans.tsx
```diff
@@ -0,0 +1,298 @@
...
+              },
+            }}
+            key={feature.name}
+          >
+            <Th
+              css={{
+                display: 'flex',
+                flexDirection: 'column',
+                justifyContent: 'center',
+                maxWidth: 335,
```
gregogun, line 244: manually setting maxWidth here (same for `Td` and `Th` elements) to adjust width according to design, as I couldn't manage to get a working solution with flex or grid

pradel (approved):
For now it's the only part where we will use the table so not creating components is fine.
If we add another table to the app, let's create a component 👍 
Good job it looks great 🚀

PR #572, opened 2022-06-28
feat: settings redesign

### LINKED ISSUE #558: ⚡ Feature | SETTING - Redesign settings's page component
We need to redesign the components of the settings page with the new design system 
Capture d’écran 2022-06-21 à 15 53 00

Figma link : https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=1919%3A12336

The inputs : https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=2239%3A2654

gregogun:
- Create reusable form components 
- Update general settings page to new design
- Add a warning to the settings page when a user hasn't saved changes

Closes #558

### DIFF
--- a/sigle/src/modules/settings/SettingsForm.tsx
+++ b/sigle/src/modules/settings/SettingsForm.tsx
@@ -41,7 +41,7 @@ const UnsavedChangesContainer = styled('div', {
 
   [`.${darkTheme} &`]: {
     boxShadow:
-      '0px 8px 20px rgba(8, 8, 8, 0.4), 0px 10px 18px rgba(8, 8, 8, 0.35), 0px 5px 14px rgba(8, 8, 8, 0.3), 0px 3px 8px rgba(8, 8, 8, 0.2), 0px 1px 5px rgba(8, 8, 8, 0.18), 0px 1px 2px rgba(8, 8, 8, 0.16), 0px 0.2px 1px rgba(8, 8, 8, 0.08)',
+      '0px 8px 20px rgba(8, 8, 8, 0.32), 0px 10px 18px rgba(8, 8, 8, 0.28), 0px 5px 14px rgba(8, 8, 8, 0.26), 0px 3px 8px rgba(8, 8, 8, 0.16), 0px 1px 5px rgba(8, 8, 8, 0.14), 0px 1px 2px rgba(8, 8, 8, 0.12), 0px 0.2px 1px rgba(8, 8, 8, 0.08)',
   },
 });
 


sigle/src/modules/settings/SettingsForm.tsx
```diff
@@ -124,6 +135,7 @@ interface SettingsFormProps {
 export const SettingsForm = ({ settings, username }: SettingsFormProps) => {
   const queryClient = useQueryClient();
   const { user } = useAuth();
+  const [unsavedChanges, setUnsavedChanges] = useState(false);
```
pradel, line 138: instead of using this react state, could we use the dirty formik prop? https://formik.org/docs/api/formik#dirty-boolean

sigle/src/modules/settings/SettingsForm.tsx
```diff
@@ -273,65 +318,66 @@ export const SettingsForm = ({ settings, username }: SettingsFormProps) => {
...
-      </StyledFormRow>
+      </FormRow>
 
-      <StyledFormRow>
+      <FormRow>
         <FormLabel>Website</FormLabel>
         <FormInput
           name="siteUrl"
           type="text"
           maxLength={50}
```
pradel, line 350: this should be updated to match the text

sigle/src/ui/Form.ts
```diff
@@ -0,0 +1,87 @@
+import { styled } from '../stitches.config';
```
pradel, line 1: you can add the export of these components in `sigle/src/ui/index.ts` so they can be imported like normal UI components :)

pradel (changes requested):

sigle/src/modules/settings/SettingsForm.tsx
gregogun: ah I wasn't aware of this property, neat thanks!

QuentinSaubadu:
1 - The save settings card should disappear after saving, otherwise we tend to think that the saving was not done properly (the button says that there are changes to save).

https://user-images.githubusercontent.com/7244762/176388214-0a7d1f8f-5996-48fe-8d01-146806aeb8db.mov

2 - The description input has a 2px stroke when active where the others have not. keep 1px stroke when you are on "active" mode.

https://user-images.githubusercontent.com/7244762/176388666-a654e7bc-33c2-4b4f-a89f-3ee51ac0abaa.mov

@gregogun

pradel (approved):

PR #575, opened 2022-06-29
fix: fix image in settings not updating after saving

gregogun:
The image is now updating when you save but there's a slight moment where it shows the previous image when the value is reset. Is there a way to prevent this from happening? 

https://user-images.githubusercontent.com/65421744/176427681-d754f75c-ac91-4cca-b198-0b3996cf10ab.mov

### DIFF
--- a/sigle/src/modules/settings/SettingsForm.tsx
+++ b/sigle/src/modules/settings/SettingsForm.tsx
@@ -199,7 +199,10 @@ export const SettingsForm = ({ settings, username }: SettingsFormProps) => {
         ...newSettings,
       });
 
-      queryClient.setQueriesData('user-settings', newSettings);
+      queryClient.setQueriesData('user-settings', {
+        ...settingsFile,
+        ...newSettings,
+      });
 
       if (customLogo) {
         setCustomLogo(undefined);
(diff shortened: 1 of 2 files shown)


sigle/src/modules/settings/SettingsForm.tsx
```diff
@@ -197,13 +199,13 @@ export const SettingsForm = ({ settings, username }: SettingsFormProps) => {
         ...newSettings,
       });
 
+      queryClient.setQueriesData('user-settings', newSettings);
```
pradel, line 202: This part should probably be 
```
      queryClient.setQueriesData('user-settings',{ ...settingsFile,
        ...newSettings,});
```
To match exactly what's saved on Gaia

sigle/src/modules/settings/SettingsForm.tsx
```diff
@@ -381,32 +383,33 @@ export const SettingsForm = ({ settings, username }: SettingsFormProps) => {
...
-            type="submit"
-            size="md"
-            color="orange"
-          >
-            {formik.isSubmitting ? 'Saving...' : 'Save changes'}
-          </Button>
-        </UnsavedChangesContainer>
-      )}
+      {formik.dirty ||
+        (customLogo && (
```
pradel, line 387: I think the correct logic here is
```suggestion
      {formik.dirty || customLogo)
         && (
```

Issue #581, opened 2022-07-04
⚡ Feature | Display Creator + tick 


QuentinSaubadu:
Users who activate the Creator + plan must have a corresponding tick next to their name on...
- The profile page (20x20px)
<img width="983" alt="Capture d’écran 2022-07-04 à 17 57 20" src="https://user-images.githubusercontent.com/7244762/177188599-4add1ed4-63eb-4f6a-a2d3-524dd4ede211.png">
FIGMA: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=3518%3A26634
- The Following and Followers page (12x12px)
<img width="839" alt="Capture d’écran 2022-07-04 à 17 58 24" src="https://user-images.githubusercontent.com/7244762/177188761-a0f26651-5f3f-474a-83c6-0179f3375ca0.png">
FIGMA: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=3640%3A30443
- The Explore page (12x12px)
<img width="809" alt="Capture d’écran 2022-07-04 à 17 59 20" src="https://user-images.githubusercontent.com/7244762/177188869-05ba20d1-c78e-4d82-8f50-1eab942ada92.png">
FIGMA: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=3580%3A24028

**The tick:** https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3070%3A6001

**SVG :** 
LIGHT MODE:
```
<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
[...]

PR #597, opened 2022-07-12
feat: show creator + badge on user profile

pradel:
Part of https://github.com/sigle/sigle/issues/581


sigle/src/modules/publicHome/components/PublicHome.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { NextSeo } from 'next-seo';
+import Image from 'next/future/image';
```
pradel, line 3: Wanted to try the new module introduced by next 12.2 👀

Issue #585, opened 2022-07-04
⚡ Feature | PROFILE - Add profile page on the left menu


QuentinSaubadu:
Add "Profile" on the left menu, bottom position.
This tab displays the profile page (old blog page).

The URL doesn't change compared to when it was a separate blog page.
<img width="947" alt="Capture d’écran 2022-07-04 à 19 15 53" src="https://user-images.githubusercontent.com/7244762/177197351-831d5da8-14eb-4468-bb34-459ca28f3580.png">

PR #609, opened 2022-07-19
feat: redesign dropdown header

### LINKED ISSUE #584: ⚡ Feature | HEADER - Redesign dropdown menu
It's time now to upgrade the dropdown menu.
FIGMA FILE: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=2355%3A3457

Before: 
Capture d’écran 2022-07-04 à 19 07 16

After:
Capture d’écran 2022-07-04 à 19 06 09

Add the username on top followed by the ID extension and the profile picture.
Clicking on it brings you to the Profile page.

FIRST SECTION :  
The default font color is now 11.
Hover and on click mode didn't change.
Add icons on the left side.

SECOND SECTION:
The default font color is now 9.
Change "Switch theme" by "Dark mode" and add a toggle on the right side: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3095%3A5547

THIRD SECTION:
Swap the red color on the logout tab for a grey 9

gregogun:
* Updates dropdown menu
* Adds profile page to dashboard

Closes #584 #585


sigle/src/ui/DropdownMenu.ts
```diff
@@ -59,7 +59,7 @@ export const DropdownMenuItem = styled(DropdownMenuPrimitive.Item, {
   display: 'flex',
   gap: '$2',
   alignItems: 'center',
-  justifyContent: 'space-between',
+  // justifyContent: 'space-between',
```
pradel, line 62: let's remove this line if not used :)

sigle/src/modules/layout/components/DashboardLayout.tsx
```diff
@@ -146,6 +148,9 @@ export const DashboardLayout = ({
     }
   };
 
+  const pathname =
+    typeof window !== 'undefined' && window.location.pathname.split('/');
```
pradel, line 152: This part is not accessible in the `router` object?

sigle/src/modules/publicHome/components/PublicHome.tsx
```diff
@@ -100,13 +101,17 @@ const PublicHomeSiteUrl = ({ siteUrl }: { siteUrl: string }) => {
...
 
-interface PublicHomeProps {
+interface ProfilePageContentProps {
   file: StoryFile;
   settings: SettingsFile;
   userInfo: { username: string; address: string };
 }
 
-export const PublicHome = ({ file, settings, userInfo }: PublicHomeProps) => {
+const ProfilePageContent = ({
```
pradel, line 110: It's really hard to review the changes made to this file, what was the reason for having a new component here? Do you think you could revert back to just one component to make the review simpler and maybe split it into another pull request if needed :)?

gregogun: Ah right yes this was done to wrap the page content in the dashboard layout if the user is visiting their own profile, otherwise not showing the dashboard layout if you are visiting another blog. But yeah sure can do that :)

pradel: It seems that with the latest change, I am now seeing the menu on another user profile.

Screenshot 2022-07-22 at 11 35 48

This is a logic you were using before.

```
{userInfo.username !== user?.username ? (
        <ProfilePageContent
          file={file}
          settings={settings}
          userInfo={userInfo}
        />
      ) : (
        <DashboardLayout>
          <ProfilePageContent
            file={file}
            settings={settings}
            userInfo={userInfo}
          />
        </DashboardLayout>
      )}
```

Another option to keep the diff clean can be to do smth like :

```
const Layout = userInfo.username !== user?.username ? React.Fragment : DashboardLayout;

<Layout>
// Code
</Layout>
```

sigle/src/modules/layout/components/DashboardLayout.tsx
pradel: As this should be shown only on the current user profile this could work no?

gregogun: Ahh yes very true, will update it 👍

sigle/src/modules/publicHome/components/PublicHome.tsx
gregogun: ohh of course! hadn't thought about this, makes a lot more sense 😅 thanks!

sigle/src/modules/layout/components/HeaderDropdown.tsx
```diff
@@ -0,0 +1,240 @@
...
+        ))}
+        <DropdownMenuSeparator />
+        {lowerNavItems.map((item) => (
+          <Link key={item.path} href={item.path} passHref>
+            <DropdownMenuItem selected={router.pathname === item.path} as="a">
+              {item.name}
+            </DropdownMenuItem>
+          </Link>
+        ))}
+        <DropdownMenuItem
```
pradel, line 226: @gregogun for the theme toggle, the switch should probably be triggered when you click on the item and not just the switch, wdyt?

sigle/src/modules/layout/components/HeaderDropdown.tsx
```diff
@@ -0,0 +1,240 @@
...
+    },
+  ];
+
+  const lowerNavItems = [
+    {
+      name: 'Settings',
+      path: '/settings',
+    },
+    {
+      name: 'Upgrade',
```
pradel, line 119: Should the upgrade button be hidden if the user does have an active creator + subscription?

Screenshot 2022-07-22 at 16 18 02

sigle/package.json
```diff
@@ -25,6 +25,7 @@
     "@radix-ui/react-hover-card": "0.1.5",
     "@radix-ui/react-icons": "1.1.0",
     "@radix-ui/react-scroll-area": "0.1.4",
+    "@radix-ui/react-switch": "0.1.5",
```
pradel, line 28: Also could you please use the v1.X as now it's released 🤓

sigle/src/modules/layout/components/HeaderDropdown.tsx
gregogun: agreed, will update it 👍

sigle/src/modules/layout/components/HeaderDropdown.tsx
```diff
@@ -0,0 +1,249 @@
...
+        ))}
+        <DropdownMenuSeparator />
+        {lowerNavItems.map((item) => (
+          <Link key={item.path} href={item.path} passHref>
+            <DropdownMenuItem selected={router.pathname === item.path} as="a">
+              {item.name}
+            </DropdownMenuItem>
+          </Link>
+        ))}
+        {!userSubscription && (
```
gregogun, line 224: now that we're doing the check here, is it worth just inlining the above element as it's now the only one in the lowerNavItems array?

pradel: Maybe smth that can be done is to push the upgrade object to the lowerNavItems item if there is no subscription, that way you don't duplicate the code and keep lowerNavItems. Does it make sense?

gregogun: okay yeah that makes sense!

sigle/src/modules/layout/components/HeaderDropdown.tsx
```diff
@@ -0,0 +1,246 @@
...
+});
+
+export const HeaderDropdown = () => {
+  const { data: settings } = useGetUserSettings();
+  const router = useRouter();
+  const { user } = useAuth();
+  const queryClient = useQueryClient();
+  const [loadingCreate, setLoadingCreate] = useState(false);
+  const { resolvedTheme, setTheme } = useTheme();
+  const { data: userSubscription } = useGetUserSubscription();
```
pradel, line 60: This will fail for legacy users, can we trigger the query only for non legacy users?

pradel (approved):
Great one 🚀

PR #613, opened 2022-07-22
feat: profile hover card

### LINKED ISSUE #604: ⚡ Feature | Dropdown profile card
Add a dropdown with the user's profile information when hovering his name on an article page.
Capture d’écran 2022-07-14 à 10 55 10

FIGMA FILE: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=2%3A925
COMPONENT: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3070%3A5931

```
/* Grey/Light/1 Background */

background: #FCFCFC;
/* SMALL LIGHT */

box-shadow: 0px 8px 20px rgba(8, 8, 8, 0.06), 0px 10px 18px rgba(8, 8, 8, 0.04), 0px 5px 14px rgba(8, 8, 8, 0.04), 0px 3px 8px rgba(8, 8, 8, 0.04), 0px 1px 5px rgba(8, 8, 8, 0.03), 0px 1px 2px rgba(8, 8, 8, 0.02), 0px 0.2px 1px rgba(8, 8, 8, 0.01);
border-radius: 8px;
```
Capture d’écran 2022-07-14 à 17 57 19

gregogun:
Adds a hover card which is displayed when hovering a writers' name on the article page.

Closes #604


sigle/src/hooks/users.ts
```diff
@@ -25,3 +25,29 @@ export const useGetUserByAddress = (
     () => UserService.getApiUsers({ userAddress: stacksAddress }),
     options
   );
+
+type GetApiUsersFollowingReturnType = Awaited<
+  ReturnType<typeof UserService.getApiUsersFollowing>
+>;
+export const useGetUsersFollowing = (
```
gregogun, line 32: I was following the methods attached to `UserService` for naming the queries and noticed this one is very similar to the `useGetUserFollowing` in the appData file - would it best to use a different naming convention in this case or do you think it's fine?

pradel:
@gregogun wdyt about changing how you approached this pr, I was thinking that this component should be a separate one. It shouldn't be aware of the view, that way we can reuse it in many places by just adding it using a few lines of code. The logic should be encapsulated and maybe the only prop is the address and / or username?

gregogun:
> @gregogun wdyt about changing how you approached this pr, I was thinking that this component should be a separate one. It shouldn't be aware of the view, that way we can reuse it in many places by just adding it using a few lines of code. The logic should be encapsulated and maybe the only prop is the address and / or username?

yeah this makes a lot of sense actually - will separate it 👍

pradel (commented):
Looks great 😍, some feedback / discussions:

1. It seems that the alignment of the text changed
Screenshot 2022-07-26 at 10 34 12
 
2. the requests made to the API to fetch the number of following/followers is made when the page load. As this data is not used directly, it would be better to only fetch it when the card is opened 

3. I find it quite weird that this modal appears only on top of the name, imho it should also appear when hovering the avatar, wdyt?

4. Should the name be clickable and redirect you to the user profile?

5. For later, should the "X following" be clickable and redirect to the following tab of the user profile, if yes we should create a separate issue to not forget?

sigle/src/hooks/users.ts
pradel: Good point, maybe we should use a different naming for the Gaia ones, eg: `useGaiaGetUserFollowing` or smth like this

gregogun:
> 

1. fixed 👍

2. yep makes sense - wdyt about the approach of using the `open` prop on the root of the hover card as a value for the `enabled` prop on the query to achieve this? 

3. hm yeah this is a good point tbh, I've relayed it to Quentin 

4. as I understand it, by default the hover card trigger is rendering `<a />` so I was following this pattern and assumed it might throw an error if no link was passed as a child - just tested removing the link and is working fine, so I could revert it to how it was working before (whole area is clickable and redirects to profile) ?

5. yeah I feel like it would be good to have this. Will create an issue for it 👍

pradel:
2. Maybe you could track if the component is opened using `onOpenChange`? That way you don't have to handle any local state.

4. Sorry, was not super clear, I meant the name inside the hover card. Also you can render anything as parent of the Hover card, it's creating a portal so won't be rendered in the `a` link directly :)

gregogun:
2. Here we would just manage local state in the ProfileCard component right? as onOpenChange is returning void - [this](https://github.com/sigle/sigle/blob/7095da1aba358057771ba7c4124bdc0ab2031031/sigle/src/modules/profileCard/ProfileCard.tsx#L96) is what I've done for it 

4. Ah right yeah that makes sense - updated :) 

Also, regarding the avatar also opening the card when hovered, I'm wondering if it's just better to wrap the entire component (avatar, name & date/reading time) in ProfileCard, or if it's fine to wrap ProfileCard component around the avatar as well?

sigle/src/modules/profileCard/ProfileCard.tsx
```diff
@@ -0,0 +1,198 @@
...
+  settings,
+  userInfo,
+}: ProfileCardProps) => {
+  const { user, isLegacy } = useAuth();
+  const { resolvedTheme } = useTheme();
+  const [isOpen, setIsOpen] = useState(false);
+  const { data: userFollowing } = useGetUserFollowing({
+    enabled: isOpen && !!user && userInfo.username !== user.username,
+  });
+  const { data: userInfoByAddress } = useGetUserByAddress(userInfo.address);
```
pradel, line 68: This one too should also be enabled only when the card is opened :)

pradel (commented):
> Also, regarding the avatar also opening the card when hovered, I'm wondering if it's just better to wrap the entire component (avatar, name & date/reading time) in ProfileCard, or if it's fine to wrap ProfileCard component around the avatar as well?

I don't think we want the date to trigger the hover card, would probably better to just write the avatar too :)

pradel (approved):
Good job 🚀 
Can you just add the changeset for this one please?

PR #618, opened 2022-07-22
feat: infinite loading explore page

pradel:


server/src/api/modules/users/explore.ts
```diff
@@ -20,24 +20,36 @@ const getUserExploreResponseSchema = {
...
           200: getUserExploreResponseSchema,
         },
       },
     },
     async (req, res) => {
+      let { page } = req.params;
+      if (!page || page === 0) {
+        page = 1;
+      }
+      // TODO set to 50 before merging
```
pradel, line 44: TODO

server/src/api/modules/users/explore.ts
```diff
@@ -2,44 +2,69 @@ import { FastifyInstance } from 'fastify';
...
-        take: 150,
+        skip: pageSize * (page - 1),
+        take: pageSize,
+        include: {
+          _count: true,
+        },
       });
-      return res.send(users);
+      const usersCount = await prisma.user.count({ where });
+      const nextPage = usersCount > pageSize * page ? page + 1 : null;
```
pradel, line 65: here we tell the client if there is more data to load or if the end of the list is reached

gregogun (approved):

PR #630, opened 2022-07-27
feat: add following and follower tabs to profile page

### LINKED ISSUE #617: ⚡ Feature | Add followers/following tabs
Users need to see their followers and the people he follows.
In order to do that, add a "Followers" tab and a "Following" tab.
Display the "Small Profile Card" (the same as the Explore page) in it : https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=2961%3A4694
Capture d’écran 2022-07-22 à 13 45 25

gregogun:
Adds following and follower tabs to the profile page.

Closes #617

### DIFF
--- a/sigle/src/hooks/appData.ts
+++ b/sigle/src/hooks/appData.ts
@@ -44,6 +44,7 @@ export const useUserFollow = () => {
       body: { stacksAddress: address, createdAt: now },
     });
     await queryClient.invalidateQueries('get-users-followers');
+    await queryClient.invalidateQueries('get-users-following');
   });
 };
 
@@ -65,6 +66,7 @@ export const useUserUnfollow = () => {
       body: { stacksAddress: address },
     });
     await queryClient.invalidateQueries('get-users-followers');
+    await queryClient.invalidateQueries('get-users-following');
   });
 };
 
--- a/sigle/src/hooks/appData.ts
+++ b/sigle/src/hooks/appData.ts
@@ -45,6 +45,7 @@ export const useUserFollow = () => {
     });
     await queryClient.invalidateQueries('get-users-followers');
     await queryClient.invalidateQueries('get-users-following');
+    await queryClient.invalidateQueries('get-user-by-address');
   });
 };
 
@@ -67,6 +68,7 @@ export const useUserUnfollow = () => {
     });
     await queryClient.invalidateQueries('get-users-followers');
     await queryClient.invalidateQueries('get-users-following');
+    await queryClient.invalidateQueries('get-user-by-address');
   });
 };
 
--- a/sigle/src/modules/profileCard/ProfileCard.tsx
+++ b/sigle/src/modules/profileCard/ProfileCard.tsx
@@ -187,12 +187,14 @@ export const ProfileCard = ({
         <Flex gap="3">
           <Link
             href={{
-              pathname: `/${userInfo.username}`,
+              pathname: `/[username]`,
               query: { tab: 'followers' },
             }}
+            as={`/${userInfo.username}`}
             passHref
           >
             <Typography
+              as="a"
               size="subheading"
               css={{ color: '$gray9', cursor: 'pointer' }}
             >
@@ -204,12 +206,14 @@ export const ProfileCard = ({
           </Link>
           <Link
             href={{
-              pathname: `/${userInfo.username}`,
+              pathname: `/[username]`,
               query: { tab: 'following' },
             }}
+            as={`/${userInfo.username}`}
             passHref
           >
             <Typography
+              as="a"
               size="subheading"
               css={{ color: '$gray9', cursor: 'pointer' }}
             >
--- a/sigle/src/modules/profileCard/ProfileCard.tsx
+++ b/sigle/src/modules/profileCard/ProfileCard.tsx
@@ -29,7 +29,6 @@ import {
 import { useTheme } from 'next-themes';
 import { sigleConfig } from '../../config';
 import { useState } from 'react';
-import { useRouter } from 'next/router';
 
 const ProfileImageContainer = styled('div', {
   display: 'flex',
--- a/sigle/src/modules/profileCard/ProfileCard.tsx
+++ b/sigle/src/modules/profileCard/ProfileCard.tsx
@@ -29,6 +29,7 @@ import {
 import { useTheme } from 'next-themes';
 import { sigleConfig } from '../../config';
 import { useState } from 'react';
+import { useRouter } from 'next/router';
 
 const ProfileImageContainer = styled('div', {
   display: 'flex',
@@ -60,6 +61,7 @@ export const ProfileCard = ({
   userInfo,
 }: ProfileCardProps) => {
   const { user, isLegacy } = useAuth();
+  const router = useRouter();
   const { resolvedTheme } = useTheme();
   const [isOpen, setIsOpen] = useState(false);
   const { data: userFollowing } = useGetGaiaUserFollowing({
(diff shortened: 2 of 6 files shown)


sigle/src/modules/userCard/UserCard.tsx
```diff
@@ -145,12 +150,12 @@ export const UserCard = ({ address, userFollowing }: UserCardProps) => {
               {isLoadingUsername ? '...' : username}
             </Typography>
           </Link>
-          {user && !isLegacy && !following && (
+          {user && user?.username !== username && !isLegacy && !following && (
```
gregogun, line 153: the username check here is hiding the button if you see your profile show up on another users' following/followers or the explore page

sigle/src/modules/userCard/UserCard.tsx
```diff
@@ -67,6 +69,9 @@ export const UserCard = ({ address, userFollowing }: UserCardProps) => {
       return names.names[0];
     }
   );
+  const { data: userFollowing } = useGetGaiaUserFollowing({
+    enabled: !!user && username !== user.username,
```
pradel, line 73: let's use data from the cache here so we don't fetch it for every list item when it's in a big list.

sigle/src/modules/publicHome/components/PublicHome.tsx
```diff
@@ -358,6 +372,16 @@ export const PublicHome = ({ file, settings, userInfo }: PublicHomeProps) => {
                 />
               ))}
             </TabsContent>
+            <TabsContent value="following">
```
pradel, line 375: regarding the approach for the tabs I think that taking another approach would be better. The content should be rendered and data fetched only when you select the tab. At the moment we fetch both the following / followers data even tho you never watch this tab. Also the URI should be a different one, that way links can redirect you to the follower tab directly etc. Wdyt ?

sigle/src/modules/userCard/UserCard.tsx
gregogun: okay sure, should we do the same as we did on the hover card and set staleTime to Infinity?

sigle/src/modules/publicHome/components/PublicHome.tsx
gregogun: ok yeah that makes sense. Regarding the URI, would [shallow routing](https://nextjs.org/docs/routing/shallow-routing) work for this?

sigle/src/modules/userCard/UserCard.tsx
pradel: Yup 👍

sigle/src/modules/publicHome/components/PublicHome.tsx
pradel: Not sure shallow routing would be the best as you would still have to fetch all the data using this approach no?

sigle/src/modules/publicHome/components/PublicHome.tsx
```diff
@@ -153,6 +169,20 @@ export const PublicHome = ({ file, settings, userInfo }: PublicHomeProps) => {
...
+    if (!value) {
+      return;
+    }
+
+    router.replace(
+      {
+        query: { ...router.query, tab: value },
+      },
+      undefined,
+      { shallow: true }
```
gregogun, line 182: prevents page reload

gregogun:
although the data for following/followers is only being fetched depending on the value from the query params, we no longer are seeing the number.
Screenshot 2022-08-01 at 09 24 25

I tried setting staleTime to Infinity similar to the profile card but still seems to show up with no followers/following if you refresh the page. Do you know how we can access the cache for each address?

pradel:
@gregogun I didn't think about this indeed, that's a problem. We need to fetch this data somehow and as there is no API endpoint to just fetch the number it's tricky. I created this pull request adding the required data, so now you should have all the data you need :) https://github.com/sigle/sigle/pull/638

gregogun:
> 

ok perfect will use that, thanks! :D

sigle/src/hooks/appData.ts
```diff
@@ -44,6 +44,7 @@ export const useUserFollow = () => {
       body: { stacksAddress: address, createdAt: now },
     });
     await queryClient.invalidateQueries('get-users-followers');
+    await queryClient.invalidateQueries('get-users-following');
```
pradel, line 47: both lines could be merged in just one invalidateQueries call to make it run faster.

sigle/src/hooks/appData.ts
```diff
@@ -65,6 +66,7 @@ export const useUserUnfollow = () => {
       body: { stacksAddress: address },
     });
     await queryClient.invalidateQueries('get-users-followers');
+    await queryClient.invalidateQueries('get-users-following');
```
pradel, line 69: same here

sigle/src/modules/profileCard/ProfileCard.tsx
```diff
@@ -185,18 +185,40 @@ export const ProfileCard = ({
...
-              {following ? following?.length : 0}
-            </Box>
-            Following
-          </Typography>
+          <Link
+            href={{
+              pathname: `/${userInfo.username}`,
+              query: { tab: 'followers' },
+            }}
+            passHref
```
pradel, line 193: the `as` prop is missing here right?

sigle/src/modules/profileCard/ProfileCard.tsx
```diff
@@ -185,18 +185,40 @@ export const ProfileCard = ({
...
+          >
+            <Typography
+              size="subheading"
+              css={{ color: '$gray9', cursor: 'pointer' }}
```
pradel, line 197: this should be an `a` tag )

sigle/src/modules/profileCard/ProfileCard.tsx
```diff
@@ -185,18 +185,40 @@ export const ProfileCard = ({
...
+              </Box>
+              Followers
+            </Typography>
+          </Link>
+          <Link
+            href={{
+              pathname: `/${userInfo.username}`,
+              query: { tab: 'following' },
+            }}
+            passHref
```
pradel, line 210: same comment here

sigle/src/modules/profileCard/ProfileCard.tsx
```diff
@@ -185,18 +185,40 @@ export const ProfileCard = ({
...
-            </Box>
-            Followers
-          </Typography>
-          <Typography size="subheading" css={{ color: '$gray9' }}>
-            <Box css={{ color: '$gray11', mr: 2 }} as="span">
-              {following ? following?.length : 0}
-            </Box>
-            Following
-          </Typography>
+          <Link
```
pradel, line 188: For this file as the API is now returning the count for the following / followers this file can be optimised and we can remove the useGetUsersFollowing hook and just use the user info. This would save us 2 API calls and data will load faster :)

sigle/src/hooks/appData.ts
pradel: Also another question, as we are using `['get-users-following', userAdress]` as the key, what exactly is this one doing? removing the cache for all the users ?

gregogun: I think this was added before the fix/update adding the key - should this be updated to include the user address?

pradel: I am not sure about this, probably the docs have the answer for this one :)

gregogun: after investigating the docs, it seems that this will invalidate any matching queries so yeah would essentially remove the cache for all the users.

Another thing I noticed is that now we are getting following/followers amount data from the userInfo object, the number is no longer updating if you follow/unfollow a user.

pradel: In that case I think the best way to fix this is to update the user info optimistically the same way we do above with get-user-following, wdyt?

gregogun: yep that makes sense 👍

regarding merging them into one invalidateQueries call, is the syntax meant to be something like this: 
`await queryClient.invalidateQueries(['get-users-followers', 'get-users-following', 'get-user-by-address'])` ? 

tried this and it is not working - also checked the docs and can't seem to find info about invalidating multiple queries with a different key. Do you know of another way or should I just call them separately?

pradel: Looks like this is only possible with the v4 of the module so let's keep it separate for now 👍

pradel (approved):

PR #646, opened 2022-08-04
feat: add bubble menu to mobile toolbar

### LINKED ISSUE #640: ⚡ Feature | EDITOR 📱 - Toolbar for mobile
In the mobile editor, we'll replace the bubble menu and the left + menu with a toolbar that we'll place on top of the keyboard.
Capture d’écran 2022-08-02 à 09 44 05

- The toolbar is scrollable (horizontally).
- We'll place a linear gradient frame on the very right side of the tool bar (background: linear-gradient(90deg, rgba(255, 255, 255, 0) 0%, #FFFFFF 100%);) to indicate the user the toolbar is scrollable.
Capture d’écran 2022-08-02 à 09 47 27

- The icons in the small button are brand new (different sizes). See: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3246%3A7743
Capture d’écran 2022-08-02 à 09 51 09

- To use the different styles, highlight some text and click on the icon you want:
Capture d’écran 2022-08-02 à 09 55 46

- The hover mode on the icon is the same as the one on the desktop

- remove the "keyboard" icon for tips, but keep the word counting.

FIGMA : https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4145%3A30407

gregogun:
Adds bubble menu to the mobile toolbar.

Closes #640

### DIFF
--- a/sigle/src/modules/editor/BubbleMenuItems.tsx
+++ b/sigle/src/modules/editor/BubbleMenuItems.tsx
@@ -9,7 +9,7 @@ import {
   StrikethroughIcon,
   UnderlineIcon,
 } from '@radix-ui/react-icons';
-import { useEffect, useState } from 'react';
+import { forwardRef, useEffect, useImperativeHandle, useState } from 'react';
 import { Flex } from '../../ui';
 
 const BubbleMenuInput = styled('input', {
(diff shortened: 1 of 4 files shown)


sigle/src/modules/editor/BubbleMenuItems.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+    const resetLink = () => {
+      setLinkState({
+        open: false,
+        value: '',
+      });
+    };
+
+    return (
+      <>
```
gregogun, line 132: With this one, as I was unable to access the tippy styles to update (element from the lib has style tag) them to show the bubble menu at the bottom of the screen on mobile, I decided to try another approach splitting the bubble items into a separate component which could be rendered in the tooltip on desktop, and at the bottom of the screen on mobile.

sigle/src/modules/editor/BubbleMenuItems.tsx
```diff
@@ -0,0 +1,195 @@
...
+  editor: Editor;
+  iconSize?: number;
+}
+
+export const BubbleMenuItems = forwardRef<ResetLink, BubbleMenuItemsProps>(
+  ({ editor, iconSize = 18 }: BubbleMenuItemsProps, ref) => {
+    const [linkState, setLinkState] = useState({ open: false, value: '' });
+
+    useImperativeHandle(ref, () => ({
+      resetLink,
```
gregogun, line 55: this along with the forwardRef is allowing the parent component `BubbleMenu` to call the `resetLink()` method when hidden. While this is working, I'm also aware the `useImperativeHandle` hook is said to be avoided in most cases. Do you know if there is a better way to achieve this, or is this use case justified?

PR #651, opened 2022-08-19
feat: mobile editor toolbar

### LINKED ISSUE #640: ⚡ Feature | EDITOR 📱 - Toolbar for mobile

### LINKED ISSUE #644: ⚡ Feature | EDITOR 📱 - Pop up paragraph menu
By clicking on the small button on the toolbar, a pop-up menu opens to let the user select the paragraph style.
Capture d’écran 2022-08-02 à 09 48 48

- The background is color 12 - opacity: 0.7; .
- The icons are the big ones (Desktop ones)
- A green (11) icon on the right side indicates to the user what is the current paragraph style. The icon name is "Dot filled big".
Capture d’écran 2022-08-03 à 13 02 06
- Choosing a paragraph style is changing the small button text and icon.
Capture d’écran 2022-08-03 à 13 05 02
- The pop-up menu has a 40 px rounded corner on top.
Capture d’écran 2022-08-03 à 13 06 53
- Choosing one of the options obviously change the paragraph style itself in the editor. (This part might be for @pradel )

### LINKED ISSUE #648: ⚡ Feature | EDITOR 📱 - Redesign modal for mobile
The mobile version of the modal should have new spaces between elements and respect the mobile ratio.

Use a color 12 (Basic colors) with a 70% opacity for the background.

Modal without metadata: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4221%3A32021

Modal with metadata: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4221%3A32239

Don't forget to **use mobile font** for the Twitter preview card.

The "edit preview" button brings you back to the settings panel.
Capture d’écran 2022-08-15 à 11 34 15

### LINKED ISSUE #649: ⚡ Feature | EDITOR 📱 - Settings panel
The settings panel needs to get a small rework to fit the mobile view.

Clicking on this icon:
Capture d’écran 2022-08-15 à 11 41 35
 opens the Settings panel.

The buttons on the bottom of the panel are in **FIXED position**

Leave 24px bottom padding under the preview card.

Please check out the Figma file: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4184%3A33176
Capture d’écran 2022-08-15 à 11 41 22

gregogun:
Closes #648 
Closes #649
Closes #644
Closes #640


sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -197,31 +271,142 @@ export const TipTapEditor = forwardRef<
     [editor]
   );
 
+  useEffect(() => {
+    window.visualViewport.addEventListener('resize', () => {
+      // detects if virtual keyboard has opened, however an imperfect solution but the best option for iOS browsers currently as it does not yet support Virtual Keyboard API
+      setSoftKeyboardIsOpen(!softKeyboardIsOpen);
+      handleViewport();
+    });
+    window.visualViewport.addEventListener('scroll', handleViewport);
```
pradel, line 280: Please remove the event listeners on unmount

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -197,31 +271,142 @@ export const TipTapEditor = forwardRef<
...
+  });
+
+  let isScrolling: ReturnType<typeof setTimeout>;
```
pradel, line 283: Here we should be using a ref to not lose the reference on rerenders

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -197,31 +271,142 @@ export const TipTapEditor = forwardRef<
...
+                <BubbleMenuItems iconSize={15} editor={editor} />
+                <Box
+                  css={{
+                    width: 2,
+                    backgroundColor: '$gray6',
+                  }}
+                />
+              </Flex>
+            )}
+            <Typography css={{ m: 0, whiteSpace: 'nowrap ' }} size="subheading">
```
pradel, line 399:
```suggestion
            <Typography css={{ m: 0, whiteSpace: 'nowrap' }} size="subheading">
```

pradel (changes requested):
Some feedback after testing:
1. **Mobile**: on a new line, I can still see the "+" menu, it should be disabled
2. **Mobile**: try to press enter many times with the keyboard open, you will see that the new line appears under the new mobile menu. It should appear on top so the user can see his writing on the new line.
3. **Desktop**: the new breakpoints seems off, now on my 14-inch I see the mobile styles for some part of the editor 🤔 
4. For the technical approach, I am wondering if having a react state value listening the window size and handling the logic based on this value would make the code cleaner. I have the feeling that some things might be better handled that way, for example, the bubble menu is still computed on mobile even tho it's hidden, it would be better to not render the react component at all. Wdyt @gregogun ?

gregogun:
> Some feedback after testing:
> 
> 1. **Mobile**: on a new line, I can still see the "+" menu, it should be disabled
> 2. **Mobile**: try to press enter many times with the keyboard open, you will see that the new line appears under the new mobile menu. It should appear on top so the user can see his writing on the new line.
> 3. **Desktop**: the new breakpoints seems off, now on my 14-inch I see the mobile styles for some part of the editor 🤔
> 4. For the technical approach, I am wondering if having a react state value listening the window size and handling the logic based on this value would make the code cleaner. I have the feeling that some things might be better handled that way, for example, the bubble menu is still computed on mobile even tho it's hidden, it would be better to not render the react component at all. Wdyt @gregogun ?

agreed, this feels like a more practical approach to solve the issues mentioned. Will use this instead 👍

gregogun:
> > 2. **Mobile**: try to press enter many times with the keyboard open, you will see that the new line appears under the new mobile menu. It should appear on top so the user can see his writing on the new line.

regarding this one, it seems the issue is with the fact that the footer is fixed but we of course need the toolbar to stick on top of the keyboard. One solution I tried was to attach a ref to the EditorContent and detect if the bottom is intersecting with the toolbar position and programmatically scroll, but I didn't manage to get a value returned from the ref.current. Just wondering if you know another way to approach this?

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -189,31 +261,145 @@ export const TipTapEditor = forwardRef<
...
   return (
     <>
-      {editor && <BubbleMenu editor={editor} />}
-      {editor && <FloatingMenu editor={editor} />}
+      {editor && !isMobile && <BubbleMenu editor={editor} />}
+      {editor && !isMobile && <FloatingMenu editor={editor} />}
 
       <StyledEditorContent editor={editor} />
-      <Container
+
```
gregogun, line 325: Extract mobile toolbar logic to separate file

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -189,31 +261,145 @@ export const TipTapEditor = forwardRef<
...
+                    gap="5"
+                  >
+                    <BubbleMenuItems iconSize={15} editor={editor} />
+                    <Box
+                      css={{
+                        width: 2,
+                        backgroundColor: '$gray6',
+                      }}
+                    />
+                    {slashCommands({ storyId: story.id })
```
gregogun, line 365: use find over filter here

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -189,31 +261,145 @@ export const TipTapEditor = forwardRef<
...
+              <>
+                <MobileFloatingMenu
+                  editor={editor}
+                  story={story}
+                  triggerDisabled={!softKeyboardIsOpen}
+                />
+                {editor && (
+                  <Flex
+                    css={{
+                      '@supports (-webkit-touch-callout: none) and (not (translate: none))':
```
gregogun, line 345: add comment here regarding ios

sigle/src/modules/editor/MobileFloatingMenu.tsx
```diff
@@ -0,0 +1,102 @@
...
+                mr: '$2',
+                br: '$1',
+
+                '& svg': {
+                  width: 15,
+                  height: 15,
+                },
+              }}
+            >
+              {activeNode(editor).icon}
```
gregogun, line 76: call once and save in a variable instead

sigle/src/modules/editor/ActiveNode.tsx
```diff
@@ -0,0 +1,56 @@
...
+  Heading3Light,
+  NumberedListLight,
+  QuoteLight,
+} from '../../icons';
+import { Editor } from '@tiptap/react';
+import { TextIcon } from '@radix-ui/react-icons';
+
+export const activeNode = (editor: Editor) => {
+  if (editor.isActive('heading', { level: 2 })) {
+    return {
```
gregogun, line 14: return properties from command list rather than hard-code

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -177,6 +220,35 @@ export const TipTapEditor = forwardRef<
       }),
     ],
     content: story.contentVersion === '2' ? story.content : '',
+    onUpdate: () => {
```
gregogun, line 223: remove this part and put into a separate pr

sigle/src/modules/editor/ActiveNode.tsx
gregogun: @pradel regarding this one, is the idea to filter the command list and look for a match with, say the title, which would ofc require manually passing it for each node, or do you know if there is a more programmatic way to do it?

sigle/src/modules/editor/EditorToolbar/EditorToolbar.tsx
```diff
@@ -0,0 +1,155 @@
...
+              css={{
+                p: 0,
+
+                '& svg': {
+                  filter: 'invert(1)',
+                },
+              }}
+              onClick={() => image.command({ editor: editor })}
+            >
+              <image.icon width={20} height={20} />
```
gregogun, line 139: @pradel do you know how we can render `Icon` here, so it looks a bit cleaner ?

sigle/src/modules/editor/extensions/SlashCommand/CommandList.tsx
```diff
@@ -110,11 +112,25 @@ export const CommandList = forwardRef<CommandListRef, CommandListProps>(
...
+            <Flex align="center">
+              <Icon width={35} height={35} />
+              <Flex direction="column" css={{ ml: '$2' }}>
+                <Typography>{title}</Typography>
+                <Typography css={{ color: '$gray9', mt: '-8px' }}>
+                  {description}
+                </Typography>
+              </Flex>
             </Flex>
+            {currentNodeName === title && (
```
gregogun, line 124: since updating the active node function to filter through the `slashCommands` which requires a storyId, we would have to pass it through here, so I thought it would be better to instead pass the node name, wdyt?

pradel: Good idea 👍

sigle/src/modules/editor/TipTapEditor.tsx
```diff
@@ -173,7 +177,7 @@ export const TipTapEditor = forwardRef<
       TipTapPlaceholder,
       // Custom extensions
       SlashCommands.configure({
-        commands: slashCommands({ storyId: story.id }),
+        commands: !isMobile ? slashCommands({ storyId: story.id }) : [],
```
pradel, line 180: here the slash command plugin is still injected in the plugin list, removing it totally is better so no code is executed :)

sigle/src/modules/editor/ActiveNode.tsx
```diff
@@ -0,0 +1,73 @@
+import { Editor } from '@tiptap/react';
+import { TextIcon } from '@radix-ui/react-icons';
+import { slashCommands } from './extensions/SlashCommand/commands';
+
+export const activeNode = (editor: Editor, storyId: string) => {
+  if (editor.isActive('heading', { level: 2 })) {
+    const bigHeading = slashCommands({ storyId }).filter(
```
pradel, line 7: Better to use .find here as we just need one item

sigle/src/modules/editor/EditorToolbar/ToolbarBubbleMenu.tsx
```diff
@@ -0,0 +1,171 @@
...
+  outline: 'none',
+});
+
+// Maybe can be used on clicks https://github.com/ueberdosis/tiptap/issues/104#issuecomment-912794709
+
+interface BubbleMenuProps {
+  editor: Editor;
+}
+
+export const ToolbarBubbleMenu = ({ editor }: BubbleMenuProps) => {
```
pradel, line 41: Wdyt of about calling this component `ToolbarMenu` instead of ToolbarBubbleMenu?

gregogun: yeah that makes more sense :)

sigle/src/modules/editor/TipTapEditor.tsx
gregogun: do you know what is the best method to only prevent it from being injected on mobile? Typescript is throwing an error if I use a conditional/`&&` operator as boolean is incompatible with the type the extension is expecting. Is there another way we can do it?

pradel: Would smth like this work?
```
    isMobile
        ? SlashCommands.configure({
            commands: slashCommands({ storyId: story.id }),
          })
        : undefined,
    ].filter((pluging) => !!plugin),
```

gregogun: Typescript is still throwing an error with this regarding undefined not being assignable :/

pradel (approved):
LGTM, @gregogun please fix the conflicts and good to go 🚀

PR #659, opened 2022-08-31
feat: add mobile header

gregogun:


gregogun:
@pradel please could you test and confirm if there is some buggy behavior if you pull the menu down slightly (so it resets position) and then pull it down again? 

it seems to work when using responsive mode on the browser but not working on my phone so just want to be sure it's not an iOS-specific bug

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -90,6 +133,41 @@ export const AppHeader = () => {
           </Box>
         </Link>
       </Flex>
+
+      <Flex gap="2">
+        <Flex
+          css={{
+            display: 'flex',
+            '@xl': {
```
pradel, line 141: @gregogun I am a bit confused with the breakpoints, but shouldn't this be `md`?

sigle/src/modules/layout/components/AppHeader.tsx
```diff
@@ -31,6 +45,31 @@ export const AppHeader = () => {
   const { resolvedTheme, setTheme } = useTheme();
   const { user, isLegacy } = useAuth();
   const { status } = useSession();
+  const [showMobileHeaderDialog, setShowMobileHeaderDialog] = useState(false);
+  const [loadingCreate, setLoadingCreate] = useState(false);
+  const router = useRouter();
+
+  const handleCreateNewPrivateStory = async () => {
```
pradel, line 52: we keep repeating this function everywhere, in the future we should think of a better way to do this, will be easy to forget some changes later on.

sigle/src/ui/Button.ts
```diff
@@ -20,9 +20,13 @@ export const Button = styled('button', {
         br: '$1',
       },
       lg: {
-        px: '$4',
-        py: '$3',
+        px: '$3',
```
pradel, line 23: @gregogun every screen will be affected by this change correct? Did you check with Quentin that other screens were okay?

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: ahh yeah that's correct it should be `md`, updated 👍

sigle/src/modules/layout/components/AppHeader.tsx
gregogun: yep that makes sense, should I create an issue for this so we don't forget?

sigle/src/ui/Button.ts
gregogun: yep confirmed this with Quentin :)

sigle/src/modules/layout/components/AppHeader.tsx
pradel: Yes please 👍 we could evaluate using smth like jotai or zustand maybe for this

sigle/src/ui/Dialog.tsx
```diff
@@ -49,20 +50,35 @@ const StyledContent = styled(DialogPrimitive.Content, {
...
   px: '$6',
   pt: 48,
   pb: '$5',
   overflow: 'scroll',
-  '@media (prefers-reduced-motion: no-preference)': {
-    animation: `${contentShow} 150ms cubic-bezier(0.16, 1, 0.3, 1)`,
-    willChange: 'transform',
+  transform: 'none',
+
+  '@xl': {
```
pradel, line 68: here should be md too no?

pradel (approved):

Issue #652, opened 2022-08-22
⚡ Feature | Drag and drop images in text container


QuentinSaubadu:
Users need to be able to drag and drop images (jpg/png/gif..) from their desktop to the editor.

When dragging an image to the editor, we'll display a thin grey (7) line of 2px between the paragraphs so the user can clearly see where he's placing the image: 
<img width="1015" alt="Capture d’écran 2022-08-22 à 17 17 32" src="https://user-images.githubusercontent.com/7244762/185958890-8b3e676f-5d42-44f6-9b18-5b9e8573ff77.png">

releasing the left mouse click will drop the image in the wanted paragraph and will load like shown
Use the spinner icon (white) we are using on the login page
Use a color 12 overlay 70% opacity to cover the image while loading: 
<img width="1009" alt="Capture d’écran 2022-08-22 à 17 17 38" src="https://user-images.githubusercontent.com/7244762/185959137-fcba87c5-9e69-4ba8-bd08-32a422408b49.png">

FIGMA FILE: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4648%3A35347

pradel:
@gregogun to handle the dropzone part, we already have the react-dropzone npm package so please use this one :)

PR #669, opened 2022-09-09
feat: handle image drop on editor

gregogun:
- Adds logic to handle drag and drop functionality on the editor
- Handles multiple images being dropped on the editor
- Adds improved loading state via a node view wrapper around image element

Fixes https://github.com/sigle/sigle/issues/652


pradel (commented):
Code lgtm ✅ 

Love the new loading component, the drag and drop experience is really smooth ❤️.

But I feel like now, it's much harder to see what's going on when you drag an existing image in the editor and move it to some block below. As the drop line color is really close to the background I couldn't figure in which block my image was about to be dropped. While comparing to the current version, I prefer a lighter color so you can really see it with the dark background. Wdyt?

gregogun:
> Code lgtm ✅
> 
> Love the new loading component, the drag and drop experience is really smooth ❤️.
> 
> But I feel like now, it's much harder to see what's going on when you drag an existing image in the editor and move it to some block below. As the drop line color is really close to the background I couldn't figure in which block my image was about to be dropped. While comparing to the current version, I prefer a lighter color so you can really see it with the dark background. Wdyt?

Spoke to Quentin about this one and increased it by a step (now at step 8), wdyt? If it's still not enough can see with him about pushing it to 9.

pradel (approved):
LGTM, once the conflicts are resolved let's merge 🚀

PR #677, opened 2022-09-15
feat: cover image drop

### LINKED ISSUE #654: ⚡ Feature | Drag and drop cover image
The drag and drop feature should also work for the cover image.

If the user drag the image on the "add cover image" container, the container itself changes into the followed design:
Capture d’écran 2022-08-22 à 18 04 50

- Add a 2px color 7 border on the container with a 8px corner radius
- Center the text (subheading 400 color 9) in the container
- Add "hand" icon

Use a fade in/out transition to make it smooth.

If the user is uploading a wrong extension file (like a gif), use the "warning" toast component we already have and write:
- **Wrong file extension. Only Jpegs and PNGs are accepted for cover images.**

Warning! The buttons also changed.
The "default muted" state stays the same (grey 9) but the font on hover mode is now 11.
Please refer to this Figma file : https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3402%3A8701
Capture d’écran 2022-08-24 à 12 49 07

gregogun:
- Improves user experience when dragging and dropping an image on the cover image trigger.

Closes #654

### DIFF
--- a/sigle/src/modules/editor/CoverImage.tsx
+++ b/sigle/src/modules/editor/CoverImage.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useState } from 'react';
 import { CameraIcon, HandIcon, TrashIcon } from '@radix-ui/react-icons';
 import { useDropzone } from 'react-dropzone';
 import { styled } from '../../stitches.config';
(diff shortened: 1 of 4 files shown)


sigle/src/modules/editor/CoverImage.tsx
```diff
@@ -49,28 +51,44 @@ export const CoverImage = ({ story, setStoryFile }: CoverImageProps) => {
...
       setLoadingSave(false);
       setStoryFile({ ...story, coverImage: coverImageUrl });
     },
     [story]
   );
-  const { getRootProps, getInputProps } = useDropzone({
+  const { getRootProps, getInputProps, fileRejections } = useDropzone({
     onDrop,
     accept: 'image/jpeg,image/png',
+    onDragEnter: () => setIsDraggingOver(true),
```
pradel, line 63: could this be replaced by `isDragActive` which is returned by `useDropzone`?

gregogun: ahh yes forgot it's exposing this, will update it now thanks :)

PR #694, opened 2022-09-28
feat: editor twitter integration

### LINKED ISSUE #672: ⚡ Feature | Editor twitter embed
## Description

Users need to embed tweets on their stories.

In order to do that, the Plus Menu has to be updated to the last version (see issue: https://github.com/sigle/sigle/issues/673)

The user needs to copy-paste a Tweet to make it work. If the user is copying something else in the input, the input is returning an error (unless for example if it is a youtube video and the command is already provided on the editor, in that case we need to detect it and change the command automatically) :
Capture d’écran 2022-09-14 à 15 38 50

When you click on the "Twitter" command in the Plus menu, an input opens in the editor in a new paragraph:

Capture d’écran 2022-09-14 à 15 39 06
Capture d’écran 2022-09-14 à 15 39 12

The embed looks like this: 

Capture d’écran 2022-09-14 à 15 39 38

Check the full component to see the different properties of the Twitter embed card: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3463%3A9689

FULL FIGMA PAGES: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4648%3A35350

**Leo's comment:** 
Behaviour:
- Handle global paste event https://github.com/ueberdosis/tiptap/blob/main/packages/extension-youtube/src/youtube.ts
- If invalid url entered into the input show error
- Saving should just remove the input if the user did not enter the link yet

gregogun:
This pr will allow users to embed tweets on their stories with the following functionality: 

- users can access the input for their tweet link via the floating menu
- if a user enters an invalid url or a broken link/deleted tweet, the correct errors will show for each edge case 
- if a user saves their story while an input is empty, it is removed (currently replaced with span tag)
- if a user pastes a link in the story that matches a valid url for a tweet, it will automatically detect and embed the tweet

## Steps to test: 
- [x] Open the floating menu and select the twitter embed item which will display the input
- [x] Paste a valid tweet url (both the tweet page url and the `Copy link to tweet` will work) and press enter. This should briefly display a loading state, then successfully generate a tweet embed
- [x] (editor only) Once the embedded tweet renders on the screen, it should create a new paragraph below it.
- [x] Save the story and refresh. The Embedded tweet should still be there
- [x] Paste an invalid url (any link that isn't a tweet url) and press enter. This should generate an error with 2 options: `Reset` and `Paste URL as a link`. 
- [x] Clicking reset should reset the input 
- [x] Clicking `Paste URL as a link` should remove the input and paste the url as plain text
- [ ] Paste a link to a deleted tweet ([can copy this one](https://twitter.com/gregogun/status/1576842362068029442?s=20&t=GapqSBdNztNGBet3ETTgFQ)) and press enter. This should show an error with the options: `Retry` and `Paste URL as a link`. 
- [ ] Clicking retry should trigger the loading state and either resolve and embed the tweet or return to the error state (in this case as the tweet is deleted it will return to error state).
- [ ] Paste a link and press save while the input is filled, then refresh the page. The filled input should still be there
- [ ] Save the story with the input empty, then refresh. The input should be removed
- [ ] Pressing backspace or delete when the input is empty should delete it
- [ ] Paste a valid tweet url as plain text and it should be automatically detected and embed the tweet
- [ ] Paste the deleted tweet as plain text and it should also be automatically detected but show the error after the loading state

Closes #672


sigle/src/modules/editor/extensions/Twitter/TwitterComponent.tsx
```diff
@@ -0,0 +1,334 @@
...
+      }
+
+      props.editor.commands.createParagraphNear();
+      setTweetCreated(true);
+      formik.setSubmitting(false);
+    });
+  };
+
+  const onKeyDown = (event: React.KeyboardEvent) => {
+    // Remove input if empty and user presses backspace or delete key
```
pradel, line 203: Great one 👍

sigle/src/modules/editor/extensions/Twitter/TwitterComponent.tsx
```diff
@@ -0,0 +1,334 @@
...
+    if (
+      (!formik.values.tweetUrl && event.key === 'Backspace') ||
+      event.key === 'Delete'
+    ) {
+      props.deleteNode();
+    }
+  };
+
+  const pasteUrlAsLink = () => {
+    const editor = props.editor;
```
pradel, line 213: here we should check that it's a valid link otherwise insert it as text

sigle/src/modules/editor/extensions/Twitter/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  addCommands() {
+    return {
+      setTweet:
+        () =>
+        ({ commands }) => {
+          commands.insertContent({
+            type: this.name,
+          });
+          commands.focus();
```
pradel, line 53: is this doing anything?

pradel (changes requested):
Some small comments and the issue shared via DM and we are good to go, good job 🚀

sigle/src/modules/editor/extensions/Twitter/index.ts
```diff
@@ -0,0 +1,100 @@
...
+  parseHTML() {
+    return [
+      {
+        tag: 'div[data-twitter]',
+      },
+    ];
+  },
+
+  renderHTML({ HTMLAttributes }) {
+    console.log({ HTMLAttributes });
```
pradel, line 88: please remove

sigle/src/modules/editor/extensions/Twitter/index.ts
gregogun: guess we don't need this as we have the solution with setTimout - removed 👍

sigle/src/modules/editor/extensions/Twitter/TwitterComponent.tsx
gregogun: we have the check for the valid link at line 133, or is there a reason we should still validate it here?

pradel: if you enter in the input any sentence which is not a link eg: "Tokenomics Research @ Stacks Foundation | Author 'Tokenomics for Builders'" and then click "Paste URL as link", it's creating an invalid link :)

gregogun: ahh yes makes sense, will update it :)

pradel (approved):

PR #708, opened 2022-10-14
feat: add fallback og image when sharing profile page

### LINKED ISSUE #684: ⚡ Feature | Display PFP in Twitter card when sharing a profile page
When sharing a writer's profile page on Twitter, we need to display the profile picture of the said writer.

How it is now:
Capture d’écran 2022-09-20 à 15 32 13

How it should be:
Capture d’écran 2022-09-20 à 16 03 53

If the writer has no PFP, we should display the Bored avatar.

gregogun:
- Adds boring avatar fallback if the user has no profile image
- Encodes image url 

Tool used to verify og meta image is working: https://www.opengraph.xyz/

Closes #684

### DIFF
--- a/sigle/src/modules/publicHome/components/PublicHome.tsx
+++ b/sigle/src/modules/publicHome/components/PublicHome.tsx
@@ -166,6 +166,7 @@ export const PublicHome = ({ file, settings, userInfo }: PublicHomeProps) => {
   const seoDescription =
     settings.siteDescription?.substring(0, 300) ||
     `Read stories from ${siteName} on Sigle, decentralised and open-source platform for Web3 writers`;
+
   const seoImage = settings.siteLogo;
 
   const Layout =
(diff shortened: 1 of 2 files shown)


sigle/src/modules/publicHome/components/PublicHome.tsx
```diff
@@ -182,6 +186,23 @@ export const PublicHome = ({ file, settings, userInfo }: PublicHomeProps) => {
     );
   };
 
+  const encodeImageUrl = (url: string | undefined) => {
+    if (!url) {
+      return;
+    }
+
+    // encode part of url that includes file name which may contain special characters
+    const split = 'settings/';
```
pradel, line 195: @gregogun why is this logic needed? `encodeURI` should not update the non special chars no?

pradel (approved):

PR #716, opened 2022-10-17
feat: update user card styles

### LINKED ISSUE #712: ⚡ Feature | 📱 Small update on small profile cards
This issue contains 2 things:

1- Updating the small profile card on mobile mode to make it more responsive like shown below

Capture d’écran 2022-10-14 à 16 53 39

FIGMA: https://www.figma.com/file/uPVMeBq0bERyViiRMXzLN8/Sigle-design-system?node-id=3816%3A7101

Be careful, pfp size changes between desktop and mobile. The follow button needs to be updated to mobile size and the biography + name stops 8px before the follow button.

**Be careful to update the card not only on the Explore page but also on the following/followers tabs on profile pages.**

2 - The small profile cards are going out of the frame on the Explore page. We need to fix that.
Capture d’écran 2022-10-14 à 17 08 46

gregogun:
Closes #712

### DIFF
--- a/sigle/src/modules/userCard/UserCard.tsx
+++ b/sigle/src/modules/userCard/UserCard.tsx
@@ -9,16 +9,31 @@ import {
   useUserUnfollow,
 } from '../../hooks/appData';
 import { styled } from '../../stitches.config';
-import { Button, Flex, Typography } from '../../ui';
+import {
+  Button,
+  Flex,
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+  Typography,
+} from '../../ui';
 import { generateAvatar } from '../../utils/boringAvatar';
 import { fetchSettings } from '../../utils/gaia/fetch';
 import { useAuth } from '../auth/AuthContext';
+import Image from 'next/future/image';
+import { useTheme } from 'next-themes';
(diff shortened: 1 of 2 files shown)


sigle/src/modules/userCard/UserCard.tsx
```diff
@@ -27,37 +43,71 @@ const ProfileImageContainer = styled('a', {
...
 interface UserCardProps {
   address: string;
 }
 
 export const UserCard = ({ address }: UserCardProps) => {
+  const { resolvedTheme } = useTheme();
   const { user, isLegacy } = useAuth();
   const { mutate: followUser } = useUserFollow();
   const { mutate: unfollowUser } = useUserUnfollow();
+  const { data: userInfoByAddress } = useGetUserByAddress(address);
```
pradel, line 110: @gregogun can we separate the badge logic to another pr? This means we will be sending 50+ requests to the server (one for each user) and will hit rate limit fast. Will need to make some server changes to get it to work fine on prod. So let's move it to another pr to merge a bit later :)

pradel (changes requested):

pradel (approved):

PR #724, opened 2022-10-18
feat: update button styles

gregogun:
This pr corrects button sizing to match the components in Figma.

### DIFF
--- a/sigle/src/ui/Button.ts
+++ b/sigle/src/ui/Button.ts
@@ -1,10 +1,11 @@
 import { styled } from '../stitches.config';
+import { Typography } from './Typography';
 
-export const Button = styled('button', {
+export const Button = styled('button', Typography, {
   display: 'inline-flex',
   justifyContent: 'center',
   alignItems: 'center',
-  fontSize: '$2',
+  fontSize: '$1',
   lineHeight: '20.4px',
   whiteSpace: 'nowrap',
 
@@ -13,12 +14,17 @@ export const Button = styled('button', {
     pointerEvents: 'none',
   },
 
+  '@xl': {
+    fontSize: '$2',
+  },
+
   variants: {
     size: {
       md: {
         px: '$3',
         py: '$1',
         br: '$1',
+        maxHeight: 28,
       },
       lg: {
         px: '$3',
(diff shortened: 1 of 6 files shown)


sigle/src/ui/Button.ts
```diff
@@ -13,12 +13,17 @@ export const Button = styled('button', {
...
+    fontSize: '$2',
+  },
+
   variants: {
     size: {
       md: {
         px: '$3',
         py: '$1',
         br: '$1',
+        maxHeight: 28,
```
pradel, line 26: if we set this max height prop here we can't have multi-line buttons anymore, so doesn't seems like the right way to fix the problem no ?

sigle/src/ui/Button.ts
```diff
@@ -13,12 +13,17 @@ export const Button = styled('button', {
     pointerEvents: 'none',
   },
 
+  '@xl': {
```
pradel, line 16: xl or md?

sigle/src/ui/Button.ts
gregogun: hmm yeah that's true - will see with Quentin if it's better to update the design then as we inspected on devtools and all properties are the same so we have no idea where the extra pixels are coming from 😅

sigle/src/ui/Button.ts
gregogun: following the breakpoints on typography which are changing at `xl` but maybe it should be `md`? as most other component styles are updating at this breakpoint. Wdyt?

pradel: I think md make more sense, xl is really for big screens

sigle/src/ui/Button.ts
pradel: that's a really strange one, not the first time I see smth like this 😓

sigle/src/ui/Button.ts
gregogun: yeah makes sense - should I also update the typography breakpoints here, or do it in a separate pr?

pradel: here sounds good 👍

gregogun:
Work for this was carried over and completed in #753

pradel:
Can we close?

gregogun:
> Can we close?

yep all good to close 👍

PR #726, opened 2022-10-19
fix: follow counters not updating

### LINKED ISSUE #682: 🐛 Bug | Follow counter not working on small profile cards
The counter on the small profile cards (when hovering a writer's name on an article) is not updating when we click on the "follow" CTA.

See screen recording here:

https://user-images.githubusercontent.com/7244762/191268470-0f2430ce-9813-4492-a1f0-a4e0cfb9f408.mov

gregogun:
Closes #682

### DIFF
--- a/.changeset/hot-olives-confess.md
+++ b/.changeset/hot-olives-confess.md
@@ -1,0 +1,5 @@
+---
+'@sigle/app': patch
+---
+
+Fix follow counter not updating after follow or unfollow action.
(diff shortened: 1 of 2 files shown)


gregogun:
@pradel With this one, for some reason, it's fixed by reverting to the way we had to run the `invalidateQueries` separately in v3. Checked the docs but couldn't find any examples of running this function with multiple query keys

PR #730, opened 2022-10-20
feat: add login modal prompt

### LINKED ISSUE #629: ⚡ Feature | EXPLORE - Add login modal on follow button
On the Explore page, when logged out, we need to display the "follow" button to convert the users.
When the logged-out user clicks on the "follow" button, display the modal below:
Capture d’écran 2022-07-26 à 09 50 04

https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=4187%3A29767

gregogun:
- Adds a login prompt on the user cards (Explore and public profile page)
- Adds login prompt to the profile cards (Public story page)

Closes #629


sigle/src/modules/loginModal/LoginModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+    'Build your community',
+    'Create your feed',
+    'Get insight analytics',
+  ];
+
+  return (
+    <Dialog open={open} onOpenChange={onClose}>
+      <DialogContent css={{ p: '$5' }} closeButton={false}>
+        <PromptImageContainer>
+          <PromptImage src="/static/img/login_to_continue_low_2.png" />
```
pradel, line 45: let's use next image component here so the image is optimised :)

sigle/src/modules/explore/Explore.tsx
```diff
@@ -11,7 +11,7 @@ export const ExploreUsers = () => {
   const { user, isLegacy } = useAuth();
   const { isLoading: isLoadingUserFollowing, data: userFollowing } =
     useGetGaiaUserFollowing({
-      enabled: !!user && !isLegacy,
+      enabled: !!user || !isLegacy,
```
pradel, line 14: can you please explain this change?

gregogun: ah yes so this is relating to the infinite load issue I mentioned when a user is logged out - it appears `!!user && isLegacy` is causing the query to remain in a loading state so I assumed it's trying to fetch which is why used the `||` operator to force it to resolve with an error.

However, after reading the docs just now I realise that `isLoading` doesn't necessarily mean a fetch is happening, so I'm wondering if it would be better to revert the change and use `fetchStatus` [here](https://github.com/sigle/sigle/blob/5c727d9558ec7901bffdf3c8a4fb71b9d127842f/sigle/src/modules/explore/Explore.tsx#L71) instead of `isLoadingUserFollowing` ? (tested it and is working 👍)

pradel: Sounds good 👍

pradel (approved):
Nice one 🚀

PR #735, opened 2022-10-24
feat: add message for legacy users on plans page

### LINKED ISSUE #717: ⚡ Feature | Add message on Plan page for Legacy
When connected with the old Legacy login, we need to add a small line of text on the Plans page saying to the user he needs to connect with the Hiro wallet to upgrade.

"This feature is available for Hiro wallet accounts only"
"Back to profile" CTA (small button) below

gregogun:
Closes #717

### DIFF
--- a/sigle/src/modules/settings/plans/CurrentPlan.tsx
+++ b/sigle/src/modules/settings/plans/CurrentPlan.tsx
@@ -72,7 +72,7 @@ export const CurrentPlan = () => {
     data: userSubscription,
   } = useGetUserSubscription();
   const [isSelectNFTDialogOpen, setIsSelectNFTDialogOpen] = useState(false);
-  const { user } = useAuth();
+  const { user, isLegacy } = useAuth();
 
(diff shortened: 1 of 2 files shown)


sigle/src/modules/settings/plans/CurrentPlan.tsx
```diff
@@ -198,6 +215,19 @@ export const CurrentPlan = () => {
         </>
       ) : null}
 
+      {!isLoading && isError && (
```
pradel, line 218: Wdyt about using the isLegacy variable injected by useAuth? isError can come from a network failure for example

pradel (changes requested):

sigle/src/modules/settings/plans/CurrentPlan.tsx
gregogun: yep that makes sense! will update it 👍

pradel (approved):

PR #744, opened 2022-11-01
feat: create dismissable flags

pradel:


server/src/dismissable-flags/dismissable-flags.service.ts
```diff
@@ -0,0 +1,56 @@
+import { Injectable } from '@nestjs/common';
+import { ApiTags } from '@nestjs/swagger';
+import { PrismaService } from '../prisma.service';
+
+export enum DismissableFlags {
```
pradel, line 5: @gregogun I think it would be better to use a string enum there, would make the code simpler

server/src/dismissable-flags/dismissable-flags.controller.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { Test, TestingModule } from '@nestjs/testing';
```
pradel, line 1: this test can be removed, will cover it with e2e later on

server/src/dismissable-flags/dismissable-flags.service.ts
gregogun: ok will update it 👍 would we need to update the enum on the client too?

gregogun:
@pradel WIP, CI fixed and conflicts resolved. Changes here were pulled into #745 which is adding the onboarding step model. Happy to close unless you want to separate the work and keep this one open?

PR #755, opened 2022-11-11
feat: add settings tabs

### LINKED ISSUE #749: ⚡ Feature | Create "Private data" page in settings panel
We need to create a "Private data" page to store user's private informations.

FIGMA: https://www.figma.com/file/elxKahbV8rj85IWOEyZeVu/Product-design?node-id=5590%3A35750

- Create a new page called "Private data" in the settings panel
- Add title, an input to enter an email and a small description
Capture d’écran 2022-11-10 à 10 55 21

If the email format is not respected, show an error message after the user clicked on "save changes":
Capture d’écran 2022-11-10 à 10 58 44

If the email format IS respected, display a checkmark icon (green 11) to notify the user the email is correctly set:
Capture d’écran 2022-11-10 à 11 01 01

### LINKED ISSUE #750: ⚡ Feature | Create "Newsletter" page in settings
Add a new page in the settings panel named "Newsletter". 

Leave it blank for now.

Capture d’écran 2022-11-10 à 11 02 48

This page will be modified after discussing with @pradel next Monday as we need to know how Mailjet is handling the private key part.

gregogun:
Closes #749 
Closes #750

### DIFF
--- a/sigle/src/modules/settings/private-data/PrivateData.tsx
+++ b/sigle/src/modules/settings/private-data/PrivateData.tsx
@@ -13,6 +13,7 @@ import {
   SwitchThumb,
   Typography,
 } from '../../../ui';
+import { isValidEmail } from '../../../utils/regex';
 import { UnsavedChanges } from '../components/UnsavedChanges';
 import { SettingsLayout } from '../SettingsLayout';
 
@@ -33,6 +34,10 @@ export const PrivateData = () => {
 
       if (values.email && values.email.length > 100) {
         errors.email = 'Name too long';
+      }
+
+      if (values.email && !isValidEmail(values.email)) {
+        errors.email = 'Invalid email, enter a new one.';
       }
 
       return errors;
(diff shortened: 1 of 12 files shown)


sigle/src/utils/regex.ts
```diff
@@ -1 +1,9 @@
 export const hexRegex = /^#([A-Fa-f0-9]{6})$/;
+
+export const isValidEmail = (email: string) => {
+  return email
```
pradel, line 4: is there a way to get the regex from next-auth somehow?

sigle/src/modules/settings/private-data/PrivateData.tsx
```diff
@@ -0,0 +1,111 @@
...
+            backgroundColor: '$gray2',
+            br: '$3',
+            mb: '$20',
+          }}
+        >
+          <Typography size="subheading">
+            Receive transactional emails and updates from Sigle
+          </Typography>
+          <Switch>
+            <SwitchThumb />
```
pradel, line 104: this should be tracked in the form state so we can send it to the server

sigle/src/utils/regex.ts
gregogun: had a look and doesn't look like there is a utility from the library that we can use to validate emails, but some alternative approaches mentioned here https://github.com/nextauthjs/next-auth/issues/1554

pradel (approved):

